PR #2, opened 2019-07-16
Add prettier and husky precommit

eminguyen:
should we use prettier or do we just want to use eslint but in either case we should add a precommit

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,5 +30,21 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "husky": "^3.0.0",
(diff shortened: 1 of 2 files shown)


leestanley:
I think adding a precommit hook with prettier should be fine. I also found this https://www.npmjs.com/package/prettier-eslint which runs prettier -> eslint --fix but that might be overkill

sumeet-bansal:
I think it'd be better to add this as a status check for PRs instead of a pre-commit hook. Since everyone's going to be branching, we only care about proper style when they're merging into master, not when they're working in their own branch, and I think it'd be annoying to have that enforced in your own branch.

PR #12, opened 2019-08-07
Set up Redux

eminguyen:
I made a temporary reducer but not sure if that's the right way of doing this. Never set up redux without any reducers before but we probably won't get to making reducers for a while


eminguyen:
Wait next time use the squash and merge button but also, the PR owner should be responsible for merging their own PRs. Remember to also change the labels (It helps a lot because it'll show up on your GitHub dashboard)

PR #21, opened 2019-08-09
Add ProfileCard (Closes #7)

### LINKED ISSUE #7: Create profile card

eminguyen:
So you can get started reviewing this PR but don't approve it yet. I'm going to write short unit testing for it first. Also, I think we also need to set up themes and reset files inside a new folder, say src/styles. Mainly because I realized we reuse a lot of the same colors and also because I don't think I should've had to overwrite margins for headers


eminguyen:
Also, realized that I should maintain the state of the progress bar so i'll get to that tonight

eminguyen:
@leestanley @coraxyc just finished this PR! can one of you review now?

PR #38, opened 2019-08-19
Circleci setup (Closes #36)

### LINKED ISSUE #36: Transition from TravisCI to CircleCI

leestanley:


sumeet-bansal:
There're some changes not related to CircleCI setup here so those should be moved  out of the PR.

Also, I'd recommend splitting up the Circle job into multiple: one for linting, one for testing, etc. so that if any fail, you can see directly in the GitHub UI what the issue is (e.g. if a branch fails a style check, you can see that the lint job failed instead of having to go to the Circle job and see which command exited with an error)&mdash;just a small ergonomics thing. See the portal API config [here](https://github.com/acmucsd/membership-portal/blob/master/.circleci/config.yml) for an example and the full Circle workflow [here](https://circleci.com/workflow-run/24e62684-20b8-415a-95a3-de9db6d79c98).

PR #67, opened 2019-08-29
Add profile card container

eminguyen:


eminguyen:
@leestanley good for review

PR #68, opened 2019-09-08
Register Form Layout (Closes #25)

### LINKED ISSUE #25: Create registration form
Should be similar to #5. Also, don't worry about giving it functionality. That'll be done in the registration page issue. This issue should mainly cover laying out the component, anywhere where you would usually put a function call (such as submitting the form) should just be replaced with a prop

leestanley:
Looks Like: https://i.imgur.com/Rq3IgoI.png

**Issues:**
- I made the "Major" field just a regular text input. I was originally going to make a dropdown of all the engineering majors, but then I realized other majors were going to join as well. I could add an autocomplete of all the majors at UCSD as a solution. 
- ~~I hard coded the "Years" options to graduating in 2020-2023. There might be a better solution to this, but it seemed the easiest at the time.~~


src/components/RegisterForm/style.less
```diff
@@ -0,0 +1,76 @@
+@import '../../styles/fonts.less';
```
eminguyen: two space indents

src/components/RegisterForm/style.less
```diff
@@ -0,0 +1,76 @@
...
+	background: @blue-2;
+	border-radius: 10px;
+	display: flex;
+	font-family: 'Roboto';
+	height: @register-height;
+	justify-content: center;
+	text-align: center;
+	width: @db-card-width;
+	label {
+		color: @white;
```
eminguyen: can you just set the color of the register card to be white or does that not work? If not, we should probably add a reset.less file

src/components/RegisterForm/style.less
```diff
@@ -0,0 +1,76 @@
...
+		line-height: 30px;
+	}
+	.formitems {
+    margin-bottom: 0px;
+    margin-top: 10px;
+		text-align: left;
+	}
+	.majoryear {
+		display: flex;
+    text-align: left;
```
eminguyen: indents

src/components/RegisterForm/style.less
```diff
@@ -0,0 +1,76 @@
...
+		text-align: left;
+		&:hover {
+			opacity: 0.9;
+		}
+	}
+	.ant-form-item-label {
+		line-height: 30px;
+	}
+	.formitems {
+    margin-bottom: 0px;
```
eminguyen: indents

src/components/RegisterForm/style.less
```diff
@@ -0,0 +1,76 @@
...
+		width: @button-width;
+		.major {
+			width: 170px;
+		}
+		.year {
+			width: 80px;
+		}
+	}
+	.names {
+		.majoryear;
```
eminguyen: why

src/components/RegisterForm/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+                name="major"
+                className="major"
+                value={props.values}
+                onChange={props.handleChange}
+                onBlur={props.handleBlur}
+              />
+            </Form.Item>
+            <Form.Item label="Year" className="formitems">
+              <Select className="year" value={props.values}>
+                <Option value="2020">2020</Option>
```
eminguyen: can you make this an array at the top and then utilize a for loop to iterate through it? By doing it that way, it'll be convenient to add more values when we have to later on.

src/components/RegisterForm/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+              Already have an account? <b>Sign in.</b>
+            </p>
+          </NavLink>
+        </form>
+      </div>
+    </div>
+  );
+};
+
+// TODO: swap out proptypes with formik's implementation
```
eminguyen: Capitalize Swap and add a period to the end

sumeet-bansal:
@leestanley If you’re going to hardcode graduation year, do it for current year through current year plus six. How would autocomplete for majors work? Would you be hardcoding that or pulling the majors from somewhere dynamically?

leestanley:
> @leestanley If you’re going to hardcode graduation year, do it for current year through current year plus six. How would autocomplete for majors work? Would you be hardcoding that or pulling the majors from somewhere dynamically?

For Ant Design's autocomplete implementation, you would need to pass in an array of all the majors. I would also need hardcode list of majors manually.

src/components/RegisterForm/style.less
leestanley: I tried targeting the register card but it didn't work

sumeet-bansal:
> > @leestanley If you’re going to hardcode graduation year, do it for current year through current year plus six. How would autocomplete for majors work? Would you be hardcoding that or pulling the majors from somewhere dynamically?
> 
> For Ant Design's autocomplete implementation, you would need to pass in an array of all the majors. I would also need hardcode list of majors manually.

Can we have an open issue for pulling majors from some UCSD source then? Can you also generate the list of graduation years dynamically instead of hardcoding them? Otherwise, we'll need to update the array yearly.

src/components/RegisterForm/index.jsx
```diff
@@ -7,9 +7,19 @@ import './style.less';
...
+  let currentYear = new Date().getFullYear();
+  let years = [];
+  for (var i = currentYear; i < currentYear + 6; i++) {
+    years.push(i.toString());
+  }
+  return years.concat(['Other']);
+}
+
+const years = getYears();
+
```
leestanley: Here was my solution to the dynamic year update

src/components/RegisterForm/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+import PropTypes from 'prop-types';
+import { Form, Input, Button, Select } from 'antd';
+import { NavLink } from 'react-router-dom';
+
+import './style.less';
+import logo from '../../assets/graphics/logo.svg';
+
+const { Option } = Select;
+
+function getYears() {
```
leestanley: Added the dynamic years

src/components/RegisterForm/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+  let currentYear = new Date().getFullYear();
+  let years = [];
+  for (var i = currentYear; i < currentYear + 6; i++) {
+    years.push(i.toString());
+  }
+  return years.concat(['Other']);
```
leestanley: Added the dynamic years

sumeet-bansal: Needs to be an int (not a string) within 6 six years of the current year (see schema [here](https://github.com/acmucsd/membership-portal/blob/master/app/db/models/user.js#L106) and example User instances [here](https://github.com/acmucsd/membership-portal/blob/master/app/db/populateDb.js)).

Also, the actual function could be a little tighter, e.g.
```javascript
function getYears() {
    const currentYear = new Date().getFullYear();
    return [...Array(6)].map((_,i) => i + currentYear);
}
```

PR #74, opened 2019-09-16
Temporary Password Reset 

leestanley:
Created a temporary password reset or else 'Forgot Password' would hyperlink to a blank page - I'll pr the log in page after this gets approved
Looks like: https://i.imgur.com/MRGIavg.png

### DIFF
--- a/src/components/PasswordForm/style.less
+++ b/src/components/PasswordForm/style.less
@@ -1,0 +1,27 @@
+@import '../../styles/fonts.less';
+@import '../../styles/vars.less';
+
+.passwordcard {
+    align-items: center;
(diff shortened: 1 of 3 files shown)


src/components/PasswordForm/style.less
```diff
@@ -0,0 +1,27 @@
...
+    flex-direction: column;
+    font-family: 'Roboto';
+    height: @password-height;
+    justify-content: center;
+    text-align: center;
+    width: @sign-in-width;
+    color: @white;
+
+    h1 {
+        color: @white;
```
eminguyen: 2 space indents

PR #77, opened 2019-09-22
Add Event CheckIn Container

### LINKED ISSUE #51: Create event check in container
Add actions so that the event check in container actually checks you into events

eminguyen:
Still needs to make a call to the backend before we can close #51


eminguyen:
> Still needs to make a call to the backend before we can close #51

@leestanley wait can you review this, the call to the backend will be in a separate PR

PR #88, opened 2019-09-24
Add requireAuth HOC

eminguyen:
Redirects user if they aren't authenticated. However, screen still shows until they're redirected. We should return a redirection screen in the meantime.

### DIFF
--- a/src/containers/requireAuth.jsx
+++ b/src/containers/requireAuth.jsx
@@ -1,0 +1,32 @@
+import React, { useEffect } from 'react';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


src/containers/requireAuth.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const mapDispatchToProps = dispatch => ({
+  redirectLogin: () => {
+    dispatch(replace('/login'));
+  },
+});
+
+const requireAuth = compose(
+  connect(mapStateToProps, mapDispatchToProps),
+  withAuth
```
ronakdev: trailing comma

PR #101, opened 2019-09-25
Add create events form

eminguyen:
Not mobile responsive but we don't even have a way to access admin on mobile yet.

### DIFF
--- a/src/components/CreateEventForm/index.jsx
+++ b/src/components/CreateEventForm/index.jsx
@@ -20,19 +20,17 @@ const months = [
   'October',
   'November',
   'December',
-]
+];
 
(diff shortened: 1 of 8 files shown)


src/components/CreateEventForm/index.jsx
```diff
@@ -0,0 +1,165 @@
...
+  'April',
+  'May',
+  'June',
+  'July',
+  'August',
+  'September',
+  'October',
+  'November',
+  'December',
+];
```
leestanley: You could also generate a list of months through this https://www.w3schools.com/jsref/jsref_getmonth.asp

PR #102, opened 2019-09-25
Connect admin form to Formik

eminguyen:
Need to also set up form validation with yup. Will do in a later PR.


src/containers/CreateEventForm.jsx
```diff
@@ -0,0 +1,28 @@
...
+      startTime: '',
+      startAm: '',
+      endTime: '',
+      endAm: '',
+      facebook: '',
+      description: '',
+    }
+  },
+  handleSubmit(values, { resetForm, props }) {
+    console.log(values)
```
leestanley: 😮

PR #104, opened 2019-09-27
Connect to registration API

eminguyen:


eminguyen:
Note that to run this you'll need to change the port since backend also uses 3000. Should we change this in .env or should we change the line in package.json?

sumeet-bansal:
In the backend's `.env`, just add a `PORT` field (see [here](https://github.com/acmucsd/membership-portal/blob/master/app/config/index.js#L8)).

PR #111, opened 2019-09-28
Add token storage (Closes #76)

### LINKED ISSUE #76: Implement auth in frontend
client will store jwt (in cookie) that will be used for requests to server

eminguyen:
Shout out to my friends at ACM @ UCLA but also we should figure out a better method of storage

### DIFF
--- a/src/utils/storage.js
+++ b/src/utils/storage.js
@@ -7,8 +7,6 @@
 const storageAvailable = (type) => {
   try {
     const storage = window[type];
-
-
     const x = '__storage_test__';
     storage.setItem(x, x);
     storage.removeItem(x);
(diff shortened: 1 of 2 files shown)


src/utils/storage.js
```diff
@@ -0,0 +1,54 @@
...
+ * @author ACM @ UCLA
+ * https://github.com/uclaacm/membership-portal-ui/blob/master/src/storage.js
+ */
+
+const storageAvailable = (type) => {
+  try {
+    const storage = window[type];
+
+
+    const x = '__storage_test__';
```
leestanley: 3 extra lines here

sumeet-bansal:
Re: better way of storage, I think the canonical way to store auth tokens is using HttpOnly cookies, but these need to be set by the server via response headers.

PR #115, opened 2019-09-28
Leaderboard api + CORS Fix

eminguyen:


src/actions/leaderboardActions.js
```diff
@@ -1,48 +1,35 @@
...
-    },
-    {
-      firstName: 'Stanley',
-      lastName: 'Lee',
-      points: 80,
-      image: 'http://placekitten.com/g/200/300',
-      rank: 'Linear Lemur',
+    const status = await response.status;
+    if (status === 401 || status === 403) {
+      // TODO: Logout the user.
```
ronakdev: since #118 has been merged, we can do this now

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "antd-demo",
   "version": "0.1.0",
   "private": true,
+  "proxy" : "http://localhost:3000",
```
sumeet-bansal: are you committing as `localhost`?

eminguyen: actually this line isn't needed, it was a fix ronak tried to fix the cors issue but it makes no difference

PR #119, opened 2019-09-28
Connect events to API

eminguyen:


src/actions/eventsActions.js
```diff
@@ -12,70 +15,68 @@ export const checkIn = (attendanceCode) => dispatch => {
...
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${Storage.get('token')}`,
+      },
+    });
 
-  dispatch({
-    type: FETCH_FUTURE_EVENTS,
-    payload: futureEvents,
-  });
+    let status = await eventsRes.status;
```
ronakdev: why is there an `await` on a variable

PR #146, opened 2019-10-20
Password Update

leestanley:
- there are no reducers yet
- security issue with passwords showing up in url
- alphabetical ordering
- the package.json npm start script should be changed back to port 8080 
- @firebase/storage is in the package json 

this is so broken 😢


src/containers/EventCheckIn.jsx
```diff
@@ -5,7 +5,7 @@ import EventCheckIn from '../components/EventCheckIn';
 import { checkIn } from '../actions/eventsActions';
 
 const EventCheckInContainer = props => {
-  const [value, setValue] = useState('');
+  const [value, setValue] = useState('');
```
eminguyen: extra space

PR #154, opened 2019-11-21
Setup Airbnb Eslint

leestanley:
this cant pass circle ci cause of all the warnings we have 😢😢😢😢😢


sumeet-bansal:
Add an npm script in `package.json` as well, e.g. the backend ([here](https://github.com/acmucsd/membership-portal/blob/master/package.json#L9)).

PR #155, opened 2019-11-23
Fix nuking of event creating form

StoneT2000:


src/containers/CreateEventForm.jsx
```diff
@@ -46,8 +46,14 @@ const FormikCreateEventForm = withFormik({
       attendanceCode: values.attendanceCode,
       description: values.description
     }
-    props.postEvent(event);
-    resetForm();
+    props.postEvent(event)
+    .then((resp) => {
+      resetForm();
+    })
+    .catch((error) => {
```
paulpan05: I don't think we need to log the error since we already notified the user in the UI

src/actions/adminActions.js
```diff
@@ -5,28 +5,32 @@ import { notify } from '../utils';
 import { logoutUser } from './authActions';
 
 export const postEvent = (event) => async dispatch => {
```
paulpan05: Does the postEvent action need to be async since we are already returning a promise that takes in an async function?

paulpan05: The postEvent function simply returns a promise object that takes in an async function. The returning part should be synchronous I think.

src/actions/adminActions.js
```diff
@@ -5,28 +5,32 @@ import { notify } from '../utils';
...
+      if (status === 401 || status === 403) {
+        dispatch(logoutUser());
+      }
 
-    const data = await response.json();
-    if (!data) throw new Error('Empty response from server');
-    if (data.error) throw new Error(data.error.message);
+      const data = await response.json();
+      if (!data) throw new Error('Empty response from server');
+      if (data.error) throw new Error(data.error.message);
```
leestanley: 2 extra spaces on indent

PR #161, opened 2020-02-12
Updates

StoneT2000:
Fix issue where portal would fetch events and once fetched, update state and cause it to fetch events again because the state updated a component, causing an endless loop of event fetches


leestanley:
lgtm

PR #165, opened 2020-02-13
Token verification

StoneT2000:
Token is verified for validity, so we now keep proper track of authentication and admin state across sessions


StoneT2000:
Don't merge until token verification is merged on backend

sumeet-bansal:
Token verification merged on the backend (see [here](https://github.com/acmucsd/membership-portal/pull/44)).

PR #166, opened 2020-02-13
Require admin auth

StoneT2000:
Admin pages to be seen only by admins now
- removed other comments
- Stopping some useEffect calls to update constantly when only needing to update once (fetchUser sometimes)


StoneT2000:
Don't merge until token verification is merged on backend

PR #178, opened 2020-03-19
Staff Check In

StoneT2000:
- Added checkbox to check off if you are staff
- box visible only to STAFF members

### DIFF
--- a/src/actions/eventsActions.js
+++ b/src/actions/eventsActions.js
@@ -14,7 +14,7 @@ import Config from '../config';
 import Storage from '../storage';
 import { notify } from '../utils';
 
-export const checkIn = (attendanceCode) => async dispatch => {
+export const checkIn = (info) => async dispatch => {
   try {
     const response = await fetch(Config.API_URL + Config.routes.attendance.attend, {
       method: 'POST',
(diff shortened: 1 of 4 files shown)


src/actions/eventsActions.js
```diff
@@ -23,7 +23,7 @@ export const checkIn = (attendanceCode) => async dispatch => {
         'Content-Type': 'application/json',
         Authorization: `Bearer ${Storage.get('token')}`,
       },
-      body: JSON.stringify({ attendanceCode }),
+      body: JSON.stringify({ attendanceCode: info.attendanceCode, asStaff: info.asStaff }),
```
sumeet-bansal: maybe use a [`pick`](https://underscorejs.org/#pick) function here? e.g. `JSON.stringify(pick(info, ['attendanceCode', 'asStaff']))`

src/components/EventCheckIn/index.jsx
```diff
@@ -20,7 +23,14 @@ const EventCheck = props => {
...
           Submit
         </Button>
+
       </div>
+      { props.user.profile.accountType === "STAFF" &&
+        <div className="staff-checkin-div">
+          <label className="staff-check-in-label">Staff Check In</label>
+          <Checkbox onChange={props.onAsStaffChange}/>
+        </div>
+      }
```
sumeet-bansal: not sure if the current format is convention but I think this block would be a little more obvious/readable as a conditional

StoneT2000: This is ok, would use a conditional if there was something to display if accountType wasn't STAFF but there isn't anything to display.

src/actions/eventsActions.js
StoneT2000: I think this works fine, using pick means we would have to import another library when there's no serious need

PR #183, opened 2020-03-23
Email verification

StoneT2000:
Front end side
- Users still register as normal, and get automatically directed to the home page of ACM after registering
- Are automatically notified each browser session on the website of lack of email verification
- Cannot access Store Page
- Are not publically ranked on leaderboard ( implemented on backend soon)


StoneT2000:
Edit, 
The resetting password also validates your email now, frontend side of things should still be the same after new push to backend PR

PR #203, opened 2020-04-20
Fix Event Points White Border

### LINKED ISSUE #202: Profile Card: White Border Around Event Points
image

steets250:
The Ant Design progress ring doesn't align cleanly within its div, so I switched it to a static ring. It's now spaced an even 5% away from the top-right corner of the event card.

Resolves #202


leestanley:
Nice first PR! I'll take a look at it in a bit.

leestanley:
LGTM 🚀

PR #204, opened 2020-04-20
Maintain Proper Past Events Order

### LINKED ISSUE #159: Events page small width screens orders past events backwards

steets250:
The PastEvents container was reversing the past events array in place on render. Rendering occurs twice on mobile browsers and this was causing the events array to flip twice and display in the wrong order. Switched to only reversing the array upon fetch so they maintain the same order for all cases with the most recent event on top.

Resolves #159


leestanley:
Going to merge in

Issue #186, opened 2020-04-06
add 'organization' field to event creation form


sumeet-bansal:

leestanley:
Closed by #213

PR #213, opened 2020-05-04
Add comittee/org field in event form

StoneT2000:
Stored as committee in database. Lets people fill in a committee part with a string

Addresses #186


StoneT2000:
sounds good

PR #214, opened 2020-05-04
Remove Adorable API for profiles

StoneT2000:
Addresses issue #212


src/utils.js
```diff
@@ -26,6 +26,16 @@ export const getDefaultProfile = () => {
   )}`;
 };
 
+/**
+ * Returns a random default profile picture out of a select few generated from getDefaultProfile
+ * @return {string} A link to the profile picture
+ */
+
+export const getDefaultProfile = () => {
+  let randomIndex = Math.floor(Math.random() * 9);
```
leestanley, line 35: why not const here?

StoneT2000: oops, changed to const

PR #233, opened 2020-05-05
Make Online Locations Clickable

### LINKED ISSUE #224: Event Card: Check if Location is a URL + Add Hyperlinking
image

Look into [this stackoverflow for help](https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url)

steets250:
Added two new helper utils and updated the event card.

Resolves #224

### DIFF
--- a/src/components/EventCard/index.jsx
+++ b/src/components/EventCard/index.jsx
@@ -10,22 +10,19 @@ import './styles.less';
 const EventCard = props => {
   const history = useHistory();
 
-  let location = <p className="location">{props.location}</p>;
-  if (isURL(props.location)) {
-    location = (
-      <a className="link" href={getAbsoluteURL(props.location)}>
-        {location}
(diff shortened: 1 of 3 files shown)


src/components/EventCard/index.jsx
```diff
@@ -3,17 +3,29 @@ import PropTypes from 'prop-types';
...
 const EventCard = props => {
   const history = useHistory();
+
+  let location = <p className="location">{props.location}</p>;
+  if (isURL(props.location)) {
+    location = (
+      <a className="link" href={getAbsoluteURL(props.location)}>
+        {location}
+      </a>
+    );
```
leestanley, line 19: why not condense this to a ternary operator inside the jsx?

steets250: Yep, much cleaner!

PR #237, opened 2020-05-05
Change Profile Link to OnClick Event

steets250:
Not sure if this is the best practice, but I used a history push rather than a link so the formatting retains Ant Design defaults for the card.

### DIFF
--- a/src/components/ProfileCard/index.jsx
+++ b/src/components/ProfileCard/index.jsx
@@ -5,33 +5,38 @@ import { Avatar, Card, Progress } from 'antd';
 
 import './style.less';
 import { getDefaultProfile } from '../../utils';
+import { useHistory } from 'react-router-dom';
 
 const ProfileCard = props => {
+  const history = useHistory();
   return (
-    <Link to="/profile">
-      <Card bordered={false} className="profile-card">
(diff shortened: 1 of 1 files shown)


leestanley:
hmm what if we wrapped <Link> around the inner elements of cards rather than around it?

steets250:
This is the content area that Ant Design creates. I'm not able to resize without having to override internal classes and making things messy.  [https://imgur.com/a/MAj7Jr5](https://imgur.com/a/MAj7Jr5) If I go back to having the link surround the card, then I'd have to copy a bunch of formatting over from Antd which is also not desirable. Forums seems to suggest onClick as the best way but I know history.push() isn't the optimal solution.

leestanley:
> This is the content area that Ant Design creates. I'm not able to resize without having to override internal classes and making things messy. https://imgur.com/a/MAj7Jr5 If I go back to having the link surround the card, then I'd have to copy a bunch of formatting over from Antd which is also not desirable. Forums seems to suggest onClick as the best way but I know history.push() isn't the optimal solution.

Alright I'll go ahead and approve it then - we can take a look back into this once we redesign the profile card. We might move away from ant design then

PR #251, opened 2020-05-09
Graduation Year update visual bug

### LINKED ISSUE #180: Can't edit graduation year on portal

StormFireFox1:
Closes #180.

Adds state set for ProfileUpdate component to reflect client-side change, as well as changing properties of component.

### DIFF
--- a/src/components/ProfileUpdate/index.jsx
+++ b/src/components/ProfileUpdate/index.jsx
@@ -10,17 +10,11 @@ import './style.less';
 const { Option } = Select;
 const { TextArea } = Input;
 
-function getYears() {
-  const currentYear = new Date().getFullYear();
-  return [...Array(6)].map((_, i) => i + currentYear);
-}
-
-const years = getYears();
+const years = [...Array(6)].map((_, i) => i + new Date().getFullYear());
 
 const ProfileUpdate = (props) => {
   const { handleBlur, handleChange, handleSubmit, user, values } = props;
 
-  const [gradYear, setGradYear] = useState('');
   const [bg, setBG] = useState(user.profile.profilePicture);
   const [fileList, setFileList] = useState([]);
   const [visible, setVisible] = useState(false);
@@ -68,9 +62,6 @@ const ProfileUpdate = (props) => {
     keys.forEach((key) => {
       props.setFieldValue(key, props.user.profile[key]);
     });
-    if (user.profile.graduationYear) {
-      setGradYear(user.profile.graduationYear);
-    }
   }, [user]);
 
   return (
@@ -154,14 +145,13 @@ const ProfileUpdate = (props) => {
           <div className="horizontal-input">
             <Form.Item label="Year">
               <Select
-                value={gradYear}
+                name="graduationYear"
+                value={values.graduationYear}
                 className="year"
                 onBlur={(value) => {
-                  setGradYear(value);
                   props.setFieldValue('graduationYear', value);
                 }}
                 onChange={(value) => {
-                  setGradYear(value);
                   props.setFieldValue('graduationYear', value);
                 }}
               >
@@ -221,6 +211,7 @@ ProfileUpdate.propTypes = {
   values: PropTypes.shape({
     firstName: PropTypes.string.isRequired,
     lastName: PropTypes.string.isRequired,
+    graduationYear: PropTypes.number.isRequired,
     major: PropTypes.string.isRequired,
     bio: PropTypes.string.isRequired,
   }).isRequired,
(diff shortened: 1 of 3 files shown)


leestanley:
This component needs some restructuring overall because there's functionality that can be pulled into the containers or the util.js - I'll do a code review at midnight

src/components/ProfileUpdate/index.jsx
```diff
@@ -62,7 +62,7 @@ const ProfileUpdate = props => {
       });
   };
   useEffect(() => {
-    let keys = ['firstName', 'lastName', 'major', 'bio'];
+    let keys = ['firstName', 'lastName', 'major', 'bio', 'graduationYear'];
```
leestanley, line 65: Is there a reason why this is a let vs. a const?

src/components/ProfileUpdate/index.jsx
```diff
@@ -142,8 +142,14 @@ const ProfileUpdate = props => {
...
                 className="year"
-                onBlur={value => props.setFieldValue('graduationYear', value)}
-                onChange={value => props.setFieldValue('graduationYear', value)}>
+                onBlur={value => {
+                  setGradYear(value);
+                  props.setFieldValue('graduationYear', value);
+                }}
+                onChange={value => {
+                  setGradYear(value);
+                  props.setFieldValue('graduationYear', value);
```
leestanley, line 151: This should be abstracted to the container with Formik, you can just follow the structure here in the container 
https://github.com/acmucsd/membership-portal-ui/blob/master/src/containers/ProfileUpdate.jsx. If you take a look at other input fields its these functions are passed down as prop as well. Typically we want to remove using anonymous functions within the jsx itself and pass these down as props or defined before the return. 

If you want to take a look at another example here is the container for RegisterForm https://github.com/acmucsd/membership-portal-ui/blob/master/src/containers/RegisterForm.jsx and the component here https://github.com/acmucsd/membership-portal-ui/blob/master/src/components/RegisterForm/index.jsx

src/components/ProfileUpdate/index.jsx
leestanley: Where is the prop type for grad year 
image

src/components/ProfileUpdate/index.jsx
StormFireFox1: ReisterForm also uses anonymous functions in the JSX, I assume you'd want both forms to have the funcs moved?

leestanley: Yea same should be done with register but we can worry about profile update for now

StormFireFox1: I've looked around, and Formik apparently cannot properly deal with Ant Design's selector. Turns out the idiomatic way to create custom Formik change handlers is to write them inline in the JSX, modifying props in the process.

I'll kick this can to the new review I requested.

src/components/ProfileUpdate/index.jsx
```diff
@@ -71,13 +58,10 @@ const ProfileUpdate = (props) => {
       });
   };
   useEffect(() => {
-    const keys = ['firstName', 'lastName', 'major', 'bio'];
+    const keys = ['firstName', 'lastName', 'major', 'bio', 'graduationYear'];
     keys.forEach((key) => {
-      setFieldValue(key, user.profile[key]);
+      props.setFieldValue(key, props.user.profile[key]);
```
leestanley, line 63: For props you can just use the destructuring above for and replace ```props.setFieldValue``` with ```setFieldValue```. You also don't need to call props for ```props.user.profile``` because it already destructured in ```const { handleBlur, handleChange, handleSubmit, user, values } = props;```

src/components/ProfileUpdate/index.jsx
```diff
@@ -161,10 +145,15 @@ const ProfileUpdate = (props) => {
...
               <Select
-                value={gradYear}
+                name="graduationYear"
+                value={values.graduationYear}
                 className="year"
-                onBlur={(value) => setFieldValue('graduationYear', value)}
-                onChange={(value) => setFieldValue('graduationYear', value)}
+                onBlur={(value) => {
+                  props.setFieldValue('graduationYear', value);
+                }}
```
leestanley, line 153: Same here

src/components/ProfileUpdate/index.jsx
```diff
@@ -161,10 +145,15 @@ const ProfileUpdate = (props) => {
...
+                onChange={(value) => {
+                  props.setFieldValue('graduationYear', value);
```
leestanley, line 155: Same here

src/components/ProfileUpdate/index.jsx
StormFireFox1: Resolved in 1fdcbdf.

src/components/ProfileUpdate/index.jsx
StormFireFox1: Resolved in 1fdcbdf.

PR #254, opened 2020-05-09
Added event handler for discard button

### LINKED ISSUE #253: Discard event button doesn't do anything

4R53N:
Resolves #253

### DIFF
--- a/src/components/EditEventForm/index.jsx
+++ b/src/components/EditEventForm/index.jsx
@@ -244,7 +244,7 @@ const EditEventForm = props => {
           <Button
             type="danger"
             onClick={() => {
-              history.goBack()
+              history.goBack();
             }}
             className="discard-button">
             Discard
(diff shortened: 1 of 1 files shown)


src/components/EditEventForm/index.jsx
```diff
@@ -240,7 +241,12 @@ const EditEventForm = props => {
           <Button type="primary" htmlType="submit" className="save-button">
             Submit Edits
           </Button>
-          <Button type="danger" className="discard-button">
+          <Button
+            type="danger"
+            onClick={() => {
+              history.goBack();
```
leestanley, line 247: Should the discard button redirect you backwards or clear the form @StoneT2000? Idk here

4R53N: The button is available for existing events too so I am not sure if clearing the form would make sense for those.

leestanley: Seems good be sure to switch your base branch to nightly

PR #255, opened 2020-05-09
Automatically Lint Staged Files on Commit

### LINKED ISSUE #227: Devops: Bring back Husky and add a commit hook for linting

steets250:
* added husky and lint staged; included in npm and yarn locks
* set up pre-commit action in package.json

Resolves #227

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,11 @@
     "customize-cra": "^0.8.0",
     "firebase": "^7.14.2",
     "formik": "^2.0.3",
+    "husky": "^4.2.5",
     "konami-code": "^0.2.1",
     "less": "^3.9.0",
     "less-loader": "^5.0.0",
+    "lint-staged": "^10.2.2",
     "lodash": "^4.17.15",
     "react": "^16.13.1",
     "react-app-rewired": "^2.1.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -12,9 +12,11 @@
     "customize-cra": "^0.8.0",
     "firebase": "^7.14.2",
     "formik": "^2.0.3",
+    "husky": "^4.2.5",
     "konami-code": "^0.2.1",
     "less": "^3.9.0",
     "less-loader": "^5.0.0",
+    "lint-staged": "^10.2.2",
```
leestanley, line 19: What are the benefits of using lint-staged versus doing it this way?
image

leestanley:
Also we should look into style lint but that'll be another issue

package.json
steets250: I was using it as a way to only check staged files rather than enforcing everything. This would be helpful if we decided to do something like a phased roll-out of Airbnb styling so that only newly changed files need to be fixed and not the entire codebase. I'll switch to the way you provided since in this case, we already know the rest of the code would pass linting.

steets250:
Oh wait, your method doesn't work. It fixes the formatting, but the unlinted code gets committed.

PR #258, opened 2020-05-10
WIP Feature delete event

### LINKED ISSUE #235: Event Card: Add delete event
image
Just in case events get cancelled/two people create the same thing

We can possibly put it down here
image

4R53N:
Resolves #235


4R53N:
Forgot to pull before adding the new branch will create a new pr after fixing it

PR #274, opened 2020-05-17
Delete event feature added

### LINKED ISSUE #235: Event Card: Add delete event

4R53N:
Resolves #235


src/containers/admin/EditEventForm.jsx
```diff
@@ -1,9 +1,30 @@
...
+        history.push('/');
+      })
+      .catch((error) => {
+        notify('Failed to delete the event', error);
+        console.log(error);
+      });
+  }
+  const propsToPass = { handleDelete, ...props };
+  if (propsToPass === {}) console.err('There must be props'); // This line is to go around eslint restrictions
+  return <EditEventform propsToPass />;
```
steets250, line 26: I believe this line should be `return <EditEventform {...propsToPass} />`.

4R53N: Oof eslint raises an error when I spread the props. Ok I will take a look into it.

src/containers/admin/EditEventForm.jsx
```diff
@@ -1,9 +1,30 @@
...
+import { getMonthIndex, notify } from '../../utils';
+import { editEvent, deleteEvent } from '../../actions/adminActions';
+
+function EditEvntFormWithDelete(props) {
+  const history = useHistory();
+  function handleDelete() {
+    props
+      .deleteEvent(props.values.uuid)
+      .then((resp) => {
+        notify(resp);
```
steets250, line 16: There's no need to notify with the response since you already call notify in the action.

src/containers/admin/EditEventForm.jsx
steets250: If it's not too many arguments, it would be best to pass them in directly. (It's definitely an annoying error though.)

steets250:
@4R53N if you can hold off on merging, the async functions are currently broken when they're taken out of Promise functions. I'm working on figuring out a fix and I'll leave a comment when I figure out what the solution is. (Same thing happens here when you had to fix the linting errors and removed the new Promise() part)

steets250:
I moved the remove event functionality to be directly inside the EditEvent component and added back the Promise object. This removes the prop spreading issue. Linting won't pass until #277 is merged. Everything else looks good to go. 👍

steets250:
I rebased with the fix, so you're all set to merge.

PR #276, opened 2020-05-18
Fix Broken Promise Requests

steets250:
When removing the asyncs inside Promise classes, some functionality broke. We'll wanna revisit the actions at a later time and determine the best use of dispatch and error handling because there's currently a mix of strategies in place.


steets250:
Oof I'm gonna try a different stragegy, linting is trying to break stuff again

PR #278, opened 2020-05-18
Add Stylelint for Less Files

### LINKED ISSUE #256: Devops: Set up stylelint
https://stylelint.io/ aylmao

steets250:
Added stylelint for style files. Applies the standard config and property alphabetical ordering. I ignored the custom font families rule since we have the font reset that applies in the event that a font isn't able to be loaded. Updated the lint and lint:fix commands to include style.

Resolves #256


steets250:
We can add more rules if you'd like, I just started out with the basic config and alphabetization.

PR #283, opened 2020-05-20
Event Card: Add event time to Upcoming cards

StormFireFox1:
Upcoming event cards display start and end time as well as the date.

Past events are omitted from this. Years are also removed from the event cards.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -110,9 +110,27 @@ export const formatDate = (time) => {
   ];
   const date = parsedDate.getDate();
   const monthIndex = parsedDate.getMonth();
-  const year = parsedDate.getFullYear();
 
-  return `${monthNames[monthIndex]} ${date} ${year}`;
+  return `${monthNames[monthIndex]} ${date}`;
+};
+/**
+ * Extracts the time from a UTC-formatted timestamp.
+ *
+ * Example: '1970-01-01T17:00:00.000Z' => '5:00 PM'
+ *
+ * @param {string} time The time in UTC string format.
+ * @return {string} The time of day.
+ */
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -110,9 +110,27 @@ export const formatDate = (time) => {
...
+ * @param {string} time The time in UTC string format.
+ * @return {string} The time of day.
+ */
+export const formatTime = (time) => {
+  const parsedTime = new Date(time);
+  const parsedHours = parsedTime.getHours();
+  const parsedMinutes = parsedTime.getMinutes();
+  const amOrPm = parsedHours >= 12 ? 'PM' : 'AM';
+  // edge case for midnight (0 in 24-hour format becomes 12 in 12-hour format)
+  const hours = parsedHours === 0 ? '12' : (parsedHours % 12).toString();
```
leestanley, line 130: Didn't check this out locally yet, but is there a reason why you pasrse this manually? I believe you can have this localized with the AM/PM appended as well with this. Check out the [SO here](https://stackoverflow.com/questions/13898423/javascript-convert-24-hour-time-of-day-string-to-12-hour-time-with-am-pm-and-no).

steets250: If we end up switching this to use the .toLocaleTimeString(), I'll switch the acmucsd.com site as well because it's relying on manual time helper methods right now.

leestanley: Oops sorry mb for not bringing that up earlier in that PR

StormFireFox1: Just so not everything breaks, I’ll use the already existing functions and replace their internals with .toLocaleTimeString().

PR #292, opened 2020-05-29
Use Dynamic Banner Sizes

steets250:
I switched it to have a mobile banner file (126kb) and regular banner file (395kb). This should help optimize site loading and the amount of data used.

### DIFF
--- a/src/components/Banner/index.jsx
+++ b/src/components/Banner/index.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import withSizes from 'react-sizes';
 
 import './style.less';
+import Config from '../../config';
 import DefaultBanner from '../../assets/graphics/default-banner.svg';
 
 const Banner = (props) => {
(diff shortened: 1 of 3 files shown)


src/components/Banner/index.jsx
```diff
@@ -13,10 +16,22 @@ const Banner = () => {
     <img
       alt="banner"
       className="banner"
-      src={Config.banner}
+      src={
+        isMobile
+          ? 'https://acmucsd.s3-us-west-1.amazonaws.com/portal/static/banner_s.png'
+          : 'https://acmucsd.s3-us-west-1.amazonaws.com/portal/static/banner_l.png'
+      }
```
leestanley, line 23: Hmm not sure if this is the right way of doing this but we could keep the AWS banner link `https://acmucsd.s3-us-west-1.amazonaws.com/portal/static/` in the config file, import it at the top, and append banner_s or banner_l in the ternary rather than keeping the full link in here. This would allow us to reuse it over and over as well. Thoughts?

steets250: That's definitely a cleaner option. I'll go ahead and switch.

steets250:
Switched to store the URL in config.

PR #300, opened 2020-06-02
Acm store

leestanley:
# Reducer Initial State
```ts
const initialState = {
  items Items[],
  cart: orderItem[],
  error: null,
};
```

Store cart order array as objects with `uuid` and `quantity`. Upon checkout render metadata through `GET /store/merchandise/:uuid`

Credits for each user can be extracted from user's state.


leestanley:
Task linked: [CU-8gk654 ACM Store](https://app.clickup.com/t/8gk654)

steets250:
Will open a new PR after codebase restructuring is complete.

PR #304, opened 2020-06-22
Typescript refactor

paulpan05:
Hopefully nothing breaks lol


src/components/DiscordPage/index.tsx
```diff
@@ -4,10 +4,11 @@ import Iframe from 'react-iframe';
 import './style.less';
 
 const DiscordPage = () => {
+  const ReactIframe = Iframe as React.ComponentClass<any>;
   return (
     <div className="discord-page">
       <h1 className="title">Discord</h1>
-      <Iframe
+      <ReactIframe
```
StormFireFox1, line 11: If I recall, the `react-iframe` package has an interface for its iframe, `IIframe`. Writing out locally, I find some fields from the properties of this component are not permitted in `react-iframe`. (`display`, `allowtransparency`, `frameborder`)

src/components/ProfileUpdate/index.tsx
```diff
@@ -71,6 +96,10 @@ const ProfileUpdate = (props) => {
     });
   }, [user]);
 
+  const InnerRefButton =  Button as React.ComponentClass<any>;
```
paulpan05, line 99: I had to type cast these as well since some fields are invalid.

src/components/ProfileUpdate/index.tsx
```diff
@@ -106,7 +135,7 @@ const ProfileUpdate = (props) => {
           ]}
         >
           <div className="upload-wrapper">
-            <Upload
+            <CustomUpload
```
paulpan05, line 138: Here as well

src/components/ProfileUpdate/index.tsx
```diff
@@ -151,14 +180,14 @@ const ProfileUpdate = (props) => {
           </Form.Item>
           <div className="horizontal-input">
             <Form.Item label="Year">
-              <Select
+              <CustomSelect
```
paulpan05, line 183: And here

src/components/EditEventForm/index.tsx
```diff
@@ -24,17 +24,34 @@ const months = [
...
+
+interface EditEventFormProps {
+  event: EventProp
+  deleteEvent: Function
+  setFieldValue: Function
+  setFieldTouched: Function
+  handleBlur: FocusEventHandler
+  handleChange: ChangeEventHandler
+  handleSubmit: FormEventHandler
+  values: {[key: string]: any}
```
StormFireFox1, line 51: Only commenting with the clarification that specific types for action payloads should (and most likely will) be added once the backend team constructs a type library to couple with the API.

src/components/LeaderListItem/index.tsx
```diff
@@ -1,12 +1,20 @@
...
 
-const LeaderListItem = (props) => {
+interface LeaderListItemProps {
+  exp: number
+  image: string
+  name: string
+  placement: number
+  uuid: string
+  rank: number
+}
```
StormFireFox1, line 15: To make it easier to reuse these interfaces, should we consider creating a separate "types" folder for all these? It might make backend's job of typing the API payloads easier, possibly.

src/components/ProfileUpdate/index.tsx
```diff
@@ -151,14 +180,14 @@ const ProfileUpdate = (props) => {
...
                 name="graduationYear"
                 value={values.graduationYear}
                 className="year"
-                onBlur={(value) => {
+                onBlur={(value: string | number) => {
                   setFieldValue('graduationYear', value);
                 }}
-                onChange={(value) => {
+                onChange={(value: string | number) => {
```
StormFireFox1, line 190: I'm wondering why `graduationYear` here is both a string and a number. I'd assume they'd be numbers all the time.

leestanley:
the god pr - i'll try to get my comments in tmmrw night

src/actions/authActions.ts
```diff
@@ -101,7 +102,7 @@ export const verifyToken = (dispatch) => async () => {
           resolve(data);
           return;
         }
-        const userData = tokenGetClaims(token);
+        const userData: {[key: string]: any} = tokenGetClaims(token);
```
StoneT2000, line 105: we should define a more defined object type for userData, store it in a types folder or smth so it can be reused throughout the website

src/components/EditEventForm/index.tsx
StoneT2000: how would we sync the types with backend though

src/components/ProfileUpdate/index.tsx
```diff
@@ -12,7 +11,33 @@ const { TextArea } = Input;
 
 const years = [...Array(6)].map((_, i) => i + new Date().getFullYear());
 
-const ProfileUpdate = (props) => {
+interface ProfileUpdateProps {
+  handleBlur: FocusEventHandler,
+  handleChange: ChangeEventHandler,
+  handleSubmit: FormEventHandler,
+  setFieldValue: Function,
+  user: {
```
StoneT2000, line 19: we really need a user type, this is being reused quite a lot

leestanley:
Resolve the comments above and double check the tsconfig/eslint config then we should be good. I'll be sure to do a second pass through.

src/components/EditEventForm/index.tsx
StormFireFox1: We could sync with backend in a couple of ways. Here's a list of all the options with no personal preference in particular:
- Monorepo the whole membership portal and keep the API model in a separate project folder. You can use TypeScript [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) to share types between the projects.
- Create a separate "types" package that can be linked to both the API and the UI with `npm link`. Then, we update types in the third repository and import them into each project's `tsconfig.json`.
- Separate "types" repository again, although this time use Git submodules instead of `npm`. Slightly less fun to implement, though.
- Installing both projects when developing for either and using some symlinking when cloning the repos to get the types synced between one another.

Monorepo probably works best for absolute complete synchronization of types, especially since our entire stack will eventually be in TS, but that's debatably harder than the separate types package.

src/components/LeaderListItem/index.tsx
```diff
@@ -6,12 +6,12 @@ import './style.less';
...
-  name: string
-  placement: number
-  uuid: string
-  rank: number
+  exp: number;
+  image: string;
+  name: string;
+  placement: number;
+  uuid: string;
+  rank: number;
```
StormFireFox1, line 14: Similarly to "users", the leaderboard rank type gets repeated a lot. We might want to abstract that as well.

Issue #265, opened 2020-05-11
Store Page: Collection Container


leestanley:

steets250:
Going to redo issues after finalizing store design.

PR #315, opened 2020-07-06
Acm Store Collection Container (Issue #265)

annsudhart:
image


src/actions/storeActions.ts
```diff
@@ -0,0 +1,45 @@
...
+    const data = await collectionsRes.json();
+
+    if (!data) throw new Error('Empty response from server');
+    if (data.error) throw new Error(data.error.message);
+    dispatch({
+        type: FETCH_COLLECTIONS,
+        payload: data.collections,
+    });
+} catch (error) {
+    notify('Unable to fetch future events!', error.message);
```
leestanley, line 39: Should this say store here?

src/components/StoreCollection/style.less
```diff
@@ -0,0 +1,21 @@
...
+
+.collection {
+
+    .collection-info {
+        margin: 15px 0px;
+    }
+
+    h2 {
+        font-size: @general-heading-2;
+        font-weight: bold;
```
leestanley, line 11: Avoid using bold - just put the font-weight at 600 I believe

src/components/StoreItemCard/style.less
```diff
@@ -0,0 +1,17 @@
+@import '../../styles/vars.less';
+
+.item-card{
+
+    border: 5px solid gray;
```
leestanley, line 5: Use `gray` your import from vars.less not the explicit name

src/actions/storeActions.ts
```diff
@@ -0,0 +1,45 @@
...
+        headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${Storage.get('token')}`,
+        },  
+    });
+
+    const { status } = collectionsRes;
+    if (status === 401 || status === 403) {
+        dispatch(logoutUser());
```
leestanley, line 26: hmm why dispatch a logout here versus a notify?

leestanley, line 26: hmm why dispatch a logout here versus a notify?

annsudhart: If they're not logged in, should we still show the store page and notify them that they're not logged in?

leestanley: Oh interesting, do we not have a protected route from acmucsd.com/{store url} that already blocks people from viewing though?

StormFireFox1: It is protected. As of now, it requires standard access, so no logged out user will see it anyway.

```jsx
            <Route
              exact
              path="/store"
              component={requireStandardAccess(requireAuth(StorePage)) as React.FC}
            />
```

See [index.tsx for routes](https://github.com/annsudhart/membership-portal-ui/blob/87b1b3e2cc6290e487f616a486a59f2bae544c4b/src/index.tsx#L87-L91).

src/actions/storeActions.ts
StormFireFox1: A more accurate error message would be welcome here. Maybe something like:

```suggestion
    notify('Unable to fetch store collections!', error.message);
```

src/components/StoreCollection/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+            discountPercentage: number;
+            itemName: string;
+            lifetimeLimit: number;
+            monthlyLimit: number;
+            picture: string;
+            price: number;
+            uuid: string;
+        }
+    ]
+}
```
StormFireFox1, line 25: We need to start thinking about that "types" library.

src/components/StorePage/index.tsx
```diff
@@ -1,17 +1,16 @@
...
 
 const StorePage: React.FC = () => {
   return (
     <div className="store-page">
-      <h1>ACM Store</h1>
-      <div className="coming-soon">
-        <Icon className="temp-icon" component={SwagIcon} />
-        <p>Coming Soon!</p>
-      </div>
+      <h1>Diamond Outfitters</h1>
```
StormFireFox1, line 12: I wonder if this is could be a dynamic property.

annsudhart: Are you talking about displaying a different element depending on the store's status (if the store has nothing for sale/can't fetch collections/under maintenance)?

src/actions/storeActions.ts
annsudhart: I'll remove the code for the dispatch then if it's already taken care of

src/components/StorePage/index.tsx
StormFireFox1: I was more or less referring to the store name changing depending on store name; for instance, maybe we want the store to be called something else after a while. It turns out, however, that this is supposed to be replaced with the store header component anyway, so it doesn't matter.

Issue #299, opened 2020-06-02
Event Card: Divider Partially Shown on Event Cards


leestanley:
### Cards with two line titles hide the divider 

![image](https://user-images.githubusercontent.com/8893880/83466924-63055980-a42d-11ea-8800-1537e2ac974d.png)

Issue #301, opened 2020-06-02
Password Reset: Wrong Error Notification


steets250:
Attempting to click submit should warn about having a blank email field rather than displaying a route issue.

<img width="457" alt="Screen Shot 2020-05-30 at 11 56 06 PM" src="https://user-images.githubusercontent.com/5706163/83477557-5e01d380-a448-11ea-974c-5df00d823129.png">

PR #327, opened 2020-08-16
Fixed email field error popup and event card css

paulpan05:


StormFireFox1:
This PR fixes #299 and #301.

src/components/EventCard/styles.less
```diff
@@ -13,80 +13,84 @@
...
 
-  img {
-    height: 150px;
-    margin: 0;
-    object-fit: cover;
-    width: 100%;
-  }
-
-  .info {
-    padding: 5% 10% 0 10%;
```
steets250, line 41: If you cut the top padding here, there should be enough room to squeeze 2-3 lines of title + include the bar at the bottom.

StormFireFox1: Removing the top padding makes the text be a bit too close to the event cover, but perhaps it's fine for a while.

StormFireFox1 (approved):
Looks good to me!

PR #334, opened 2020-08-22
Indicate Attended Events

### LINKED ISSUE #332: Home Page: Mark Attended Events
Events should have some indication that a member has checked in to them.

steets250:
Marks attended events with a green point ring. <s>Waiting on [this backend PR](https://github.com/acmucsd/membership-portal/pull/111) before merging.</s> Edit: PR merged, now ready for review.

Resolves #332


src/containers/PastEvents.tsx
```diff
@@ -22,20 +29,23 @@ interface PastEventsContainerProps {
...
 }
 
 const PastEventsContainer: React.FC<PastEventsContainerProps> = (props) => {
-  const { auth, events } = props;
+  const { auth, events, attendance } = props;
 
   useEffect(() => {
     props.fetchPastEvents();
+    props.fetchAttendance();
   }, []);
```
leestanley, line 42: Is there a reason why the functions here can be destructured and have to be called with the props?

steets250: @leestanley There's an issue because the functions are imported at the top of the file, so if the variables were destructured, they would be seen as being redeclared even though the functions passed through props have been connected to redux.

PR #353, opened 2020-09-16
Suborg dropdown in CreateEventForm

StormFireFox1:
This PR adds a drop-down to the "Committee" entry of the Create Event form.

Additionally, this PR also fixes an API route change for attendances; the middleware
needed modification to account for a property change in the response of the
`GET api/v1/attendance` route.

### DIFF
--- a/src/components/admin/CreateEventForm/index.tsx
+++ b/src/components/admin/CreateEventForm/index.tsx
@@ -31,6 +31,8 @@ for (let i = 1; i <= 12; i += 1) {
   hours.push(i);
 }
 
+const suborgs = ['General', 'AI', 'Cyber', 'Design', 'Hack', 'Innovate'];
+
 interface CreateEventFormProps {
   handleBlur: FocusEventHandler;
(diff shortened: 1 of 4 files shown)


src/components/admin/CreateEventForm/index.tsx
```diff
@@ -31,6 +31,8 @@ for (let i = 1; i <= 12; i += 1) {
   hours.push(i);
 }
 
+const suborgs = ['General', 'AI', 'Cyber', 'Design', 'Hack', 'Innovate'];
+
```
leestanley, line 35: Why not pass this as a prop down from container

leestanley (changes requested):

src/components/admin/CreateEventForm/index.tsx
StormFireFox1: The other properties are in the same file as well. For now, I'd say it's fine?

leestanley (approved):

PR #357, opened 2020-09-19
Mini-Redesign Leaderboard Page

### LINKED ISSUE #348: Leaderboard Page: Mini Portal Redesign

annsudhart:
Closes #348.

List of changes made:
- Card design for top three is new, and cards are space-between instead of space-around justified
- Font is all DM Sans
- Leader List Item is more compact (76px height -> 70px height) and has bolded text except for level
- For large screens, I made it so the page doesn't exceed 1700px width

image

image


src/components/Leaderboard/style.less
```diff
@@ -1,13 +1,16 @@
 .leaderboard {
+  font-family: 'DM Sans', sans-serif;
```
steets250, line 2: This line isn't necessary, as we set the sitewide font to be DM Sans.

steets250 (dismissed):
Overall looks good, but on mobile, the first and second place items have no padding between them:
Screen Shot 2020-09-19 at 7 59 21 PM

Also, please switch the PR to merge into `mini-redesign` so that the font changes take effect.

steets250 (commented):
I cleaned up an issue with the font being added twice. The one lingering issue is that the top 3 arrange a bit weirdly on narrower screens. 

Current: 
Screen Shot 2020-09-19 at 11 30 24 PM

Previous:
Screen Shot 2020-09-19 at 11 30 39 PM

Are you able to retain the behavior of the previous style while also doing the space-between? I'm going to add @leestanley as a reviewer for this PR cause I think he'll have better judgment on what to go with.

leestanley:
> I cleaned up an issue with the font being added twice. The one lingering issue is that the top 3 arrange a bit weirdly on narrower screens.
> 
> Current:
> Screen Shot 2020-09-19 at 11 30 24 PM
> 
> Previous:
> Screen Shot 2020-09-19 at 11 30 39 PM
> 
> Are you able to retain the behavior of the previous style while also doing the space-between? I'm going to add @leestanley as a reviewer for this PR cause I think he'll have better judgment on what to go with.

Hmmm I agree with Steven here - space-between does look a lot better on this screen size. Although on desktop I like how it spans the entire div. Is it possible that we can get the best of both worlds by just adding a media query here?

leestanley (approved):

PR #359, opened 2020-09-27
Mini Redesign: Update Sidebar

### LINKED ISSUE #347: Sidebar: Portal Mini Redesign

steets250:
Resolves #347


src/components/NavBarVertical/style.less
```diff
@@ -3,29 +3,21 @@
 
 .navbar-vertical {
   border-right: 1px solid rgba(51, 51, 51, 0.2);
+  color: #333;
```
annsudhart, line 6: You can use the `@black` variable in https://github.com/acmucsd/membership-portal-ui/blob/4f2a620c01ae56152661f58d9063fd589ab120e5/src/styles/vars.less#L11

leestanley:
image
- Can this just be a fixed div? When you scroll, the sidebar icons disappear so I feel like keeping them static would be better. 

- Can we also add the feedback and logout on the bottom as well like the Figma. For the feedback form just link it to `acmurl.com/portal-feedback`. I'll make the gform after this gets merged
image

PR #363, opened 2020-09-28
pull ucsd majors for autocomplete

StoneT2000:
Screen Shot 2020-09-28 at 1 53 09 PM
major change is making major field longer for majors with longer names

custom major names are not possible anymore, instead users can put in "other"


leestanley:
image
The div isn't tall enough on my screen to fit the ACM logo and has very little padding on the `Already have an account? Sign in.`

src/components/RegisterForm/majors.json
```diff
@@ -0,0 +1,3 @@
...
...dies — Philosophy","Italian Studies","Japanese Studies","Jewish Studies","Latin American Studies","Latin American Studies with a Concentration in Mexico","Latin American Studies with a Concentration in Migration and Border Studies","Language Studies","Linguistics","Literature/ Writing","Literatures in English","Spanish Literature","World Literature and Culture","Mathematics","Mathematics (Applied)","Mathematics — Applied Science","Mathematics — Computer Science","Probability & Statistics","Aerospace Engineering","Environmental Engineering","Mechanical Engineering","Interdisciplinary Computing and the Arts","Music","Music Humanities","Nanoengineering","Philosophy","General Physics","Physics","Physics-Biophysics","Political Science","Psychology","Business Psychology","Study of Religion","Russian, East European and Eurasian Studies","Earth Sciences","Marine Biology","Oceanic and Atmospheric Sciences","Sociology","Structural Engineering","Dance","Theatre","Third World Studies","Undeclared","Undeclared — Humanities/ Arts","Undeclared — Physical Sciences","Undeclared — Social Sciences","Urban Studies and Planning","Real Estate and Development","Visual Arts","Speculative Design", "Other"]
```
leestanley, line 2: Can we do some more condensing here by moving all the specializations of `International Studies: xyz` to just `International Studies`. Same with Undeclared/Bioengineering

leestanley (changes requested):

leestanley:
Not part of this PR but not sure why the year defaults to `0` here 👀 
image

steets250:
It defaults to 0 cause the default value of that field isn't set. If we assign the first year in the array, then the drop down will work correctly. Idk if you want to include that in this PR, but it's like a 1-2 line fix.

steets250:
Brushing the dust off this one... can you add a hint to the field so it says "Major" before a major is selected, and then switch the major field to being a full width field? It was fine with the old design but the redesign shrunk the width of the field and now the names get cut off:

Screen Shot 2020-12-17 at 5 11 23 PM

seanye24:
Moved year field to new line, added major placeholder
#363

steets250 (approved):
@seanye24  looks good! You can go ahead and merge.

PR #370, opened 2020-10-05
Add lazy-loading to Event Card images

StormFireFox1:
All images on Event Cards are now lazy-loaded using external package.

This change bumps the average load time of home page from 7.29 seconds down to
1.94 seconds, which is a massive difference. Here are screenshots from the
Firefox Developer Tools showing the difference in loading times before the
lazy-load change:

Load time before lazy-loading.

And after it:

Load time after lazy loading

While the images still require loading when rendered, the initial loading shock
is not as large. However, important to note that the actual image sizes should
be reduced down from 1080p, as that resolution is wasteful for the size of the
images used.

The resolution change is either mandated by the backend, or can be done with an
external script uploading old event cards to an S3 bucket at a downscaled
resolution and updating existing events using it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14893,6 +14893,15 @@
         "prop-types": "^15.5.8"
       }
     },
+    "react-lazy-load-image-component": {
+      "version": "1.5.0",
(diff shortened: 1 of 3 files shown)


steets250 (approved):
Thanks Matei, very cool!

PR #379, opened 2020-10-18
Event Administration Updates

### LINKED ISSUE #350: Create Event: Datetime Fixes + Multiday

steets250:
Added new date + time fields for event creation and editing. Added client-side validation to ensure all fields are filled in. Skipping use of Formik + Yum validation for now as it would require rewriting the form layout.

Resolves #350

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "less": "^3.9.0",
     "less-loader": "^5.0.0",
     "lodash": "^4.17.15",
+    "moment": "^2.29.1",
     "prop-types": "^15.6.2",
     "react": "^16.13.1",
     "react-app-rewired": "^2.1.3",
--- a/src/components/admin/EditEventForm/index.tsx
+++ b/src/components/admin/EditEventForm/index.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, FocusEventHandler, ChangeEventHandler, FormEventHandl
 import { Form, Input, Button, Select, DatePicker, TimePicker } from 'antd';
 import { useParams, useHistory } from 'react-router-dom';
 import * as moment from 'moment';
-import { notify } from '../../utils';
+import { notify } from '../../../utils';
 
 import './style.less';
 
--- a/src/components/admin/EditEventForm/index.tsx
+++ b/src/components/admin/EditEventForm/index.tsx
@@ -135,12 +135,14 @@ const EditEventForm: React.FC<EditEventFormProps> = (props) => {
           <div className="horizontal-input">
             <Form.Item className="date-wrapper" label="Start Date">
               <DatePicker
+                className="date"
                 value={values.startDate}
                 onChange={(date) => setFieldValue('startDate', date)}
               />
             </Form.Item>
             <Form.Item className="time-wrapper" label="Start Time">
               <TimePicker
+                className="time"
                 use12Hours
                 format="h:mm a"
                 minuteStep={15}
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "less": "^3.9.0",
     "less-loader": "^5.0.0",
     "lodash": "^4.17.15",
+    "moment": "^2.29.1",
```
StormFireFox1, line 20: Moment.js is deprecated (see [here](https://momentjs.com/docs/#/-project-status/)) and it's probably best to use other alternatives for libraries, since who knows what will happen. You may want to try:
- https://moment.github.io/luxon/ (straight from the same people)
- https://day.js.org/

src/components/admin/EditEventForm/index.tsx
```diff
@@ -141,14 +105,24 @@ const EditEventForm: React.FC<EditEventFormProps> = (props) => {
...
+              onBlur={() => setFieldTouched('committee', true)}
+              filterOption={(input, option) => {
+                const suborg: string = option.props.children as string;
+                return suborg.toLowerCase().indexOf(input.toLowerCase()) >= 0;
+              }}
+            >
+              {suborgs.map((suborg) => (
+                <Option value={`${suborg}`}>{suborg}</Option>
+              ))}
+            </Select>
```
StormFireFox1, line 125: Purely a nitpick, but you _could_ add a `Co-authored-by: StormFireFox1 <matei@gard.us>` to your commit here since this is technically from my other PR (which admittedly was not 100% complete).

This is 100% just me ranting about idiomatic Git practices, so evidently a nitpick in our case.

StormFireFox1 (changes requested):
One serious change and one nitpick, at least on my part.

package.json
steets250: The problem is that Ant Design `v3.x`, which we currently use, has Moment.js as a built-in dependency, so switching to using a different date package wouldn't have any effect. If there ends up being an issue, we can upgrade to Ant Design 4.0 during the redesign and make the switch.

src/components/admin/EditEventForm/index.tsx
steets250: If you show me how, I'd be happy to give you credit for the corresponding commit / for this PR.

StormFireFox1: It's too late now, but the idea is you add a line of this format:

`Co-authored-by: GitHub Username <email@example.com>`

In the commit message of the commit you are crediting for the co-author. If you want to, you can amend that commit with an interactive rebase and force a push, but that's too much effort for a minimal thing. It's fine now, but useful note for the future.

There are others, too, such as:
```
Signed-off-by: # for legal purposes, check the Developer Certificate of Origin on Google
Reported-by: # for bugfixes, crediting the bug reporter
Co-authored-by
Improved-by
Acked-by: # mention in commit or content itself was acknowledged
Tested-by
Reviewed-by
```

package.json
StormFireFox1: Got it. Will make an issue for this.

StormFireFox1 (approved):
Looks good to me!

src/containers/admin/CreateEventForm.tsx
```diff
@@ -2,63 +2,95 @@ import { connect } from 'react-redux';
...
+
+    if (values.attendanceCode === '') {
+      notify('Event Creation Error', 'Attendance Code is required.');
+      return;
+    }
+
+    if (values.committee === '') {
+      notify('Event Creation Error', 'Community is required.');
+      return;
     }
```
leestanley, line 80: Hmm I wonder if there's a more elegant way for error checking

leestanley (approved):

src/containers/admin/CreateEventForm.tsx
steets250: There is, using Formik form validation, but I was trying to not totally rewrite the pages yet. These can definitely be cleaned up in the rewrite.

PR #389, opened 2020-10-28
Oopsie Woopsie

steets250:
Set the wrong version for `react-router` and busted the site. Site is now unbusted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "react-infinite-scroller": "^1.2.4",
     "react-lazy-load-image-component": "^1.5.0",
     "react-redux": "^7.1.0",
-    "react-router": "^4.3.1",
+    "react-router": "^5.0.0",
     "react-router-dom": "^5.1.2",
     "react-scripts": "3.2.0",
     "react-sizes": "^2.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "react-infinite-scroller": "^1.2.4",
     "react-lazy-load-image-component": "^1.5.0",
     "react-redux": "^7.1.0",
-    "react-router": "^4.3.1",
+    "react-router": "^5.0.0",
```
StormFireFox1, line 31: Is this line really necessary? I don't recall us having it before.

StormFireFox1 (commented):
I'm confused a bit, but otherwise I'll take your word on the un-busting.

package.json
steets250: It was a missing dependency. If I didn't include, it would pop up every time we ran yarn.

StormFireFox1: Hmm. Likely `npm` popped it in without being explicitly requested?

steets250: Screen Shot 2020-10-27 at 8 33 39 PM
`npm` doesn't warn on install, but those packages have been missing.

leestanley (approved):

Issue #384, opened 2020-10-24
Leaderboard Page: Add time-based window, increase loading count


leestanley:
- [ ] Started at end of school year, e.g. 2019-2020 school year = June 15th 2019 - > June 15th 2020
- [x] Increase loading count to 100

annsudhart:
@leestanley The current Figma design shows different timeframes by quarter, are we just doing 
- 2020-2021 school year (9/21/20-Current)
- 2019-2020 school year (6/15/19-6/15/20)
- All Time?

![image](https://user-images.githubusercontent.com/32887362/97659965-91638100-1a2d-11eb-886a-09d37c210e93.png)

leestanley:
> @leestanley The current Figma design shows different timeframes by quarter, are we just doing
> 
> * 2020-2021 school year (9/21/20-Current)
> * 2019-2020 school year (6/15/19-6/15/20)
> * All Time?
> 
> ![image](https://user-images.githubusercontent.com/32887362/97659965-91638100-1a2d-11eb-886a-09d37c210e93.png)

Prefer the new redesign with a dropdown

![image](https://user-images.githubusercontent.com/8893880/97660471-e8b62100-1a2e-11eb-9793-afd510e81723.png)

PR #395, opened 2020-12-05
Added sliding leaderboard

annsudhart:
Resolves issue #384


src/containers/FourAndMore.tsx
```diff
@@ -40,23 +43,81 @@ const getFourAndMore = (users: { [key: string]: any }) => {
...
+      text: '2020-2021',
+      start: 1592204400, // June 15, 2020 in epoch seconds
+      end: 1623740400, // June 15, 2021 in epoch seconds
+    },
+    {
+      text: '2019-2020',
+      start: 1560582000, // June 15, 2019 in epoch seconds
+      end: 1592204400, // June 15, 2020 in epoch seconds
+    },
+  ];
```
leestanley, line 87: Wonder if we can make this dynamic for upcoming years so that we don't have to update this array in future iterations. Possibly pulling the current year -> incrementing year by 1.

annsudhart: That's doable - we can have a variable for the earliest year (`earliest = 1560582000, // June 15, 2019 in epoch seconds`), then increment by `yearInSeconds = 31540000` for each school year

src/containers/FourAndMore.tsx
```diff
@@ -40,23 +43,84 @@ const getFourAndMore = (users: { [key: string]: any }) => {
...
-const LIMIT = 20;
+const LIMIT = 100;
 const FourAndMoreContainer: React.FC<FourAndMoreContainerProps> = (props) => {
   const { users } = props;
   const [page, setPage] = useState(0);
   const [prevUserLength, setPrevUserLength] = useState(0);
+  const [timeframe, setTimeframe] = useState('All Time');
+  const [startTime, setStartTime] = useState(0);
+  const [endTime, setEndTime] = useState(0);
+  const earliestDate = 1560582000; // June 15, 2019 in epoch seconds
```
sumeet-bansal, line 54:
```suggestion
  const earliestDate = new Date('Jun 15, 2019').valueOf();
```

steets250:
The sliding window should also apply to the TopThree component.

annsudhart:
Shoot, I completely overlooked that

steets250 (approved):

PR #402, opened 2021-01-13
Don't unnecessarily reverse past events

sumeet-bansal:
Past events endpoint now defaults to reversing the order, so no need to reverse here.


sumeet-bansal:
The corresponding backend PR (acmucsd/membership-portal#149) hasn't merged yet.

steets250:
Great. I'll review once the backend PR is merged in.

trulyronak (approved):
i'd recommend just changing the request to send in the reverse flag

Issue #414, opened 2021-02-10
Leaderboard Limit Issue


steets250:
There's an issue where the leaderboard isn't refreshing / setting correctly once it hits the loading limit.

<img width="950" alt="Screen_Shot_2021-02-10_at_12 51 47_PM" src="https://user-images.githubusercontent.com/5706163/107575069-474d7800-6ba4-11eb-87ea-888fd2e9222e.png">

PR #415, opened 2021-02-11
Added clear leaderboard function

annsudhart:
Should solve issue #414


annsudhart:
Had a mixup with the branches, will use `andrea/sliding-leaderboard` 😅

PR #416, opened 2021-02-21
Added generic api handler

### LINKED ISSUE #406: API: Create a generic function to interact with the API

seanye24:
Resolves #406 

- adds ```fetchService``` to ```utils.ts``` as generic api handler
- adds type declaration file ```types.ts``` with relevant ```fetchService``` types
- replaces old fetch code in action files with ```fetchService```


steets250 (approved):
This is awesome! I looked through and tested out most of the routes, so I think you're good to merge.

PR #417, opened 2021-03-08
Checkin Code Encode + Decoding

steets250:


trulyronak (approved):
looks good to me!

trulyronak (changes requested):
Doesn't work — window.location is an object

trulyronak (approved):
muy bueno now

PR #420, opened 2021-03-29
Trim whitespace for "Location" in Event forms

StormFireFox1:
URL's are not rendered in the Event card due to the internal "Location"
value occasionally containing trailing whitespace and tripping the regex
for URL's. Trim whitespace on "Location" value before submitting a
"Create Event" or "Edit Event" form.


steets250 (approved):
Thanks Matei, very cool!

PR #426, opened 2021-04-05
Lowercase detected URL's in Event forms

### LINKED ISSUE #424: Convert URLs to Lowercase
If an event location is a URL, it should be converted to all lowercase before being sent to the backend. While there may technically be cases where the URL capitalization matters [here for more info](https://stackoverflow.com/questions/7996919/should-url-be-case-sensitive), I don't think we'd ever encounter that since we only use acmurl.com and zoom.us links.

unknown

StormFireFox1:
URL's may occasionally contained mixed case, which
is not perfectly visualized. Ensure any detected URL's
are all made lowercase.

Resolves #424.


steets250 (approved):
Thanks Matei, very cool!

PR #427, opened 2021-04-05
Allow not modifying cover when editing event

StormFireFox1:
New Yup schema does not allow not modifying the cover upon submitting an
Edit Event form. Add nullable constraint to Yup schema and selectively
assign cover to event object depending on nullness.


steets250 (approved):
Thanks for taking care of this. Eventually we'll want to display the current cover image so it doesn't look blank. I'll add a separate issue for that.

PR #429, opened 2021-04-11
Added Clear Leaderboard

annsudhart:
Should solve issue #414


steets250 (approved):
Hopefully clear leaderboard does the trick, I know it can be hard to test when not on production due to the lack of users.

PR #437, opened 2021-04-19
Static majors list

### LINKED ISSUE #403: Edit Profile: Add static majors list
User registration uses a static major list, but the edit profile component still allows for manual entry of a major.

seanye24:
* resolves #403 
* add constants directory


StormFireFox1:
Looked over this, it seems good; I think we could store the major data into `assets`, since that's typically where most static data goes, and I'm sure this JSON may be useful to others who'd be interested in acquiring the data.

seanye24:
> Looked over this, it seems good; I think we could store the major data into `assets`, since that's typically where most static data goes, and I'm sure this JSON may be useful to others who'd be interested in acquiring the data.

I was planning on doing that, but I wasn't sure if `assets` was reserved more for media/styling stuff.

steets250 (approved):
Looks good, I tested with both a valid and invalid starting value and it worked either way.

PR #439, opened 2021-04-19
Reset password notification

### LINKED ISSUE #374: Resetting Password should throw a Notification

seanye24:
resolves #374 

Notifications seem to work already, the only issue was that the page refresh on form submit wasn't being handled.


StormFireFox1 (approved):
Nice catch!

PR #441, opened 2021-04-25
Admin retro add functionality

annsudhart:
image

Note that the discard button doesn't work - not just for this page, but for all admin functions. I can work to fix that and add it to this PR

Edit: Discard button works now!


src/admin/adminActions.ts
```diff
@@ -129,6 +129,57 @@ export const awardPoints: ThunkActionCreator = (pointDetails: any) => async (dis
...
+      return;
+    }
+    try {
+      const url = `${Config.API_URL}${Config.routes.admin.attendance}`;
+      await fetchService(url, 'POST', 'json', {
+        requiresAuthorization: true,
+        payload: JSON.stringify({
+          users: attendanceDetails.attendees,
+          event: attendanceDetails.event,
+          asStaff: true,
```
steets250, line 151: Is `asStaff` supposed to be `true` here? I'm pretty sure that's only used for when someone checks into an event as a staff member, which wouldn't apply in this use case.

annsudhart: Oh, I thought it was to check if the person submitting the retro add was staff! I'll fix it and add a checkmark for staff adding

steets250:
@annsudhart  Looks great so far. Would you mind switching the event input field to a dropdown with a list of past events? Right now, the only way to know the event UUID is to know to check the URL when editing that event. I think it would be a cleaner experience to just pick the event from a list.

annsudhart:
@steets250 Should be good for review!

steets250 (approved):
Looks great. I tweaked the event selector so that it was a dropdown rather than a text field. Can you double check that it works? I tested it and everything seemed to be good. If so, feel free to merge in.

annsudhart:
It works ✔

PR #442, opened 2021-04-27
fix: prevent bio from horizontally overflowing off of page

duanwilliam:
image

idk


StormFireFox1 (approved):
Nice fix! Thanks for the catch.

steets250 (approved):
Looks good for now, we will tweak later on when we add a character limit / resize the bio display.

PR #444, opened 2021-04-29
Add even more ranks

StormFireFox1:
Cap for ranks hit yet again. Add even more ranks to joke about it.


steets250 (approved):
This is fine for now, but let's hit up membership to see if they wanna make new ranks.

steets250:
Denied by membership. Nice try bruv.

PR #450, opened 2021-05-23
Event cover resizing

### LINKED ISSUE #425: Admin: Image Resize Ability
Add the ability to resize an event cover to the correct ratio (16:9) and pixel size (960x540). Update the ratio of the event cards to reflect this size, as right now, they are slightly off the correct ratio.

seanye24:
resolves #425 

Uses `antd-img-crop` to provide 16:9 cropping modal for admin event cover uploads.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "dependencies": {
     "@ant-design/icons": "^4.0.6",
     "antd": "3.24.3",
+    "antd-img-crop": "^3.14.1",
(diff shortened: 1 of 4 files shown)


steets250 (approved):
Looks great. I changed the destination branch from `master` to `event-covers`, because I want to make some other changes to the cover upload interface, so I'll work off of your changes before committing everything to master.

PR #451, opened 2021-05-25
Fix Delete Event Logic

### LINKED ISSUE #447: Admin: Display Notification when Deleting Event

steets250:
The logic for checking if an event was deleted was outdated since the number of deleted events is no longer sent by the backend. In addition, the notification for that particular route was duplicated between the api function and the actual api call.

Resolves #447.


StormFireFox1 (approved):
This looks fine, though we might want to somehow tie `reject` to `notify` so this doesn't incidentally occur in the future.

PR #453, opened 2021-05-26
Clip long URL's in EventCard field

StormFireFox1:
Long URL's or locations overflow off of the event card and
do not display. Clip them by overflowing them with an ellipsis.


steets250 (approved):
Tested and working on long URLs.

PR #455, opened 2021-05-31
Past events Time Filter

annsudhart:
Used the JS filter function to filter events. I also made the timeframe variable into a var that can be accessed in the utils folder

image

### DIFF
--- a/src/event/containers/PastEvents.tsx
+++ b/src/event/containers/PastEvents.tsx
@@ -1,11 +1,13 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
+import { Menu, Dropdown } from 'antd';
 
 import EventCard from '../components/EventCard';
 import EventsList from '../components/EventsList';
 import background from '../../assets/graphics/background.svg';
+import { ReactComponent as ArrowsIcon } from '../../assets/icons/caret-icon-double.svg';
 import { fetchAttendance as fetchAttendanceConnect, fetchPastEvents as fetchPastEventsConnect } from '../eventActions';
-import { formatDate } from '../../utils';
+import { formatDate, timeframeData } from '../../utils';
 
 interface PastEventsContainerProps {
   attendance: [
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -173,3 +173,23 @@ export const fetchService = async (url: string, requestMethod: HttpRequestMethod
...
+    text: '2020-2021',
+    start: earliestDate + yearInSeconds,
+    end: earliestDate + yearInSeconds * 2,
+  },
+  {
+    text: '2019-2020',
+    start: earliestDate,
+    end: earliestDate + yearInSeconds,
+  },
+];
```
StormFireFox1, line 195: For a lot of these "timeframe" items, we could consider using BreadBot's timeframes for school quarters and academic years, pulled directly from UCSD's Academic Calendar:

https://github.com/acmucsd/discord-bot/blob/master/src/assets/quartersAndYears.json

src/event/containers/PastEvents.tsx
```diff
@@ -44,32 +46,75 @@ interface PastEventsContainerProps {
 const PastEventsContainer: React.FC<PastEventsContainerProps> = (props) => {
   const { auth, events, attendance, fetchAttendance, fetchPastEvents } = props;
 
+  const [timeframe, setTimeframe] = useState('All Time');
+  const [shownEvents, setShownEvents] = useState<any[]>(events);
```
StormFireFox1, line 50: You can define Event as an interface and use it instead of `any[]`, though that would require adding it before the Props, probably. However, types might be better left off in the future, when the membership portal API has its types package exposed.

src/layout/components/HomePage/style.less
```diff
@@ -38,6 +38,24 @@
...
+    width: fit-content;
+
+    svg {
+      margin-left: 5px;
+      vertical-align: middle;
+    }
+  }
+
+  .ant-dropdown-menu-item {
+    width: fit-content !important;
```
StormFireFox1, line 57: General question: did we ever decide whether using `!important` in the portal stylesheets was allowed or not? I don't recall.

StormFireFox1 (commented):
Looks pretty good! Left a few nits that warrant some extra discussion, but I'd merge this after discussing these.

steets250 (changes requested):
Dropdown functionality looks good, but there were a couple issues when testing:

1. I can't view events 2019-2020 as the plugin doesn't seem to have that year. Can the date range in that plugin stretch a couple years back in case some earlier quarters need to be referenced?
2. When looking at the deploy preview, I'm unable to view the example past events unless "All Time" is selected.
3. Is there an easy way we could have the dropdown only display an option if it has events in it? Options like 2028-2029 aren't going to be needed for a while.

StormFireFox1:
> I can't view events 2019-2020 as the plugin doesn't seem to have that year. Can the date range in that plugin stretch a couple years back in case some earlier quarters need to be referenced?

Added to `ucsd-quarters-years@1.1.0`.

> When looking at the deploy preview, I'm unable to view the example past events unless "All Time" is selected.

That's unfortunately expected behavior; if an event does not fit in an academic year, it's not considered for filtering. I've adjusted the filter process to "bunch up" events occuring outside an academic year to the previous academic year.

> Is there an easy way we could have the dropdown only display an option if it has events in it? Options like 2028-2029 aren't going to be needed for a while.

I fixed this.

steets250 (approved):
Looks good to me, thanks for addressing those final comments.

PR #463, opened 2021-08-01
Add Collection Item Card

### LINKED ISSUE #457: Collection Item Card
**Name(s)**
`CollectionItemCard`

**Description**
This is the item card that is displayed as part of a larger collection. 

**Image**
Screen Shot 2021-07-05 at 12 25 55 PM

**Figma**
https://www.figma.com/file/7i9MK6r5km22gynYlqC0Uj/DesignxPortal?node-id=226%3A6504

**Input Data**
https://www.notion.so/acmucsd/Merch-Store-Specification-Spring-2021-d0d5fa4d16de472aa7e13534e060f8ec#c7653e2d1daa4c85990e98edd375c316

**State**
This should be a static component with no reliance on state / dispatching.

**Notes**
- If `hidden` is true, return null.
- When clicked (regardless of whether it's in stock), it should link to: `/item/:uuid`.
- It should perform a check of the stock for each option, if all options are out of stock, the image should be gray and should display "Out of Stock" in red instead of the price.
- If all prices are the same, display a single price. Otherwise, display a range from min to max price.
- If any of the options are on sale, display a sale price with a strikethrough of the original price, and display the sale price in red. If multiple options are on sale, display the cheapest one.

StormFireFox1:
Resolves #457.

Adds the `CollectionItemCard` component, which uses no state or dispatching.

Includes test data for items. These should be replaced with backend's test data once a Collection component is made, which should fetch the items' data.

This fits the requested specifications, with the additional check for options of 0% discount being cheaper than any discounted options on an item. This is handled in a clear manner by adding red "discount" text without strikethroughs for the cheapest option.

**Screenshot of items**

Screencap of item cards using all the specified varieties.


src/store/components/CollectionItemCard/index.tsx
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+
+import './style.less';
+import { useHistory } from 'react-router';
```
steets250, line 4: Let's group the library imports before local file imports.

src/store/components/CollectionItemCard/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+type Uuid = string;
```
steets250, line 6: Types should go in `src/types.ts` for now. If we get too many, we can eventually split them into the existing folder structure.

src/store/components/CollectionItemCard/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  };
+
+  return (
+    <div className="item-card" onClick={() => history.push(`/store/item/${uuid}`)}>
+      <div className={outOfStock ? 'out-of-stock item-icon' : 'item-icon'}>
+        <img src={picture} alt={description} className={outOfStock ? 'transparent item-image' : 'item-image'} />
+      </div>
+      <h2>{itemName}</h2>
+      <div className="item-information">
+        <svg width="18" height="17" viewBox="0 0 18 17" fill="none" xmlns="http://www.w3.org/2000/svg">
```
steets250, line 111: Let's pull this `svg` out into `src/assets/icons` in the event that the icon is reused in different components.

src/store/components/StorePage/index.tsx
```diff
@@ -1,16 +1,121 @@
...
 const StorePage: React.FC = () => {
   return (
     <div className="store-page">
       <h1>ACM Store</h1>
-      <div className="coming-soon">
-        <Icon className="temp-icon" component={SwagIcon} />
-        <p>Coming Fall 2021!</p>
+      <div className="item-card-holder">
+        {items.map((item) => (
+          <CollectionItemCard
```
steets250, line 110: Can `item` be passed directly if you give the items array the type `MerchandiseItemModel[]`? It seems odd to spread the props individually since the property names match exactly.

steets250 (changes requested):
Looks great! The price/sale handling logic is clean and overall code is written well. I left a couple of comments to resolve, and please change the destination branch to `store` so we don't end up with store changes on the live site.

src/store/components/CollectionItemCard/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  const outOfStock = options.every((option) => option.quantity === 0);
+  const onSale = options.some((option) => option.discountPercentage !== 0);
+  const priceRange = options.reduce(
+    (acc, option) => {
+      const newAcc = acc;
+      if (acc.low > option.price) {
+        newAcc.low = option.price;
+      } else if (acc.high < option.price) {
```
steets250, line 52: This line should check if the `option.price` is higher **and** lower than the current ranges. Otherwise, single option items don't have a correct `high` value.

```suggestion
      }
      if (acc.high < option.price) {
```

src/store/components/StorePage/index.tsx
StormFireFox1: I figured that it's possible the API responses may not match our current model 1:1, so I deliberately passed the correct props here. For instance, items may also have collection as part of their property when received from the DB, but we don't want that property passed down to the component as well.
That being said, if it continues to remain a 1:1 pass-through, then I'll change this later.

src/store/components/CollectionItemCard/index.tsx
StormFireFox1: Nice catch. My bad.

src/store/components/StorePage/index.tsx
steets250: This is fine, and will be taken care of by the CollectionPage component. I went through all code and didn't notice that this was more of a demo rather than part of the actual component.

steets250:
One more thing: for price displays, can you add `.toLocaleString('en-US')` when displaying prices? Design has it stylized with commas for numbers > 999, so I want to follow that pattern here as well.

steets250 (approved):
LGTM 🥳

PR #471, opened 2021-09-28
Add undeclared major

steets250:


StormFireFox1 (approved):
I don't know man, I think we need to discuss the position of the name in the file. Perhaps it should be first.

`/s`

PR #478, opened 2021-10-15
The ACM Store (TM)

steets250:


duanwilliam:
it's not member facing so ig its lower priority but the create collection page validates the form on each render, which includes typing in the input bc the input is controlled, so you get spammed with form errors when typing in the name of the new collection because the description and theme colors are empty

image

steets250:
Thanks @duanwilliam, we'll look into that issue and make sure validation is present for the other forms as well.

trulyronak (approved):
This looks like it should work, what's the worst that could happen?

PR #481, opened 2021-12-01
Cart Restructure

steets250:
The store cart now contains objects with the structure `{ item: PublicMerchItem; optionUUID: string; quantity: number; }`, so that the `CartDisplay` component can pull the item data locally, rather than needing to call an API route to retrieve the cart.


steets250:
P.S. plz ignore the linting error, I specifically left an empty style block in the store starter code, so that will be resolved once more of the components get filled out.

StormFireFox1 (approved):
Looks great!

PR #482, opened 2021-12-06
Cart updates

yesean:
This PR refactors `CartItem` and `state.store.cart` (Redux store) to hold `PublicMerchItemOption` and all of the relevant metadata instead of the just the option UUID. Storing option UUIDs made sense when used in conjunction with the [`getCart` endpoint ](https://github.com/acmucsd/membership-portal/pull/209) but since we're storing the entire cart metadata on the client-side, it makes more sense to fill the Redux store with that metadata as well. It also makes it more convenient for the `CartDisplay` as the option metadata doesn't need to be unnecessarily searched for.

The store reducer and client dispatcher (`ItemPage`) have been updated to reflect these changes. `OptionSelector` has also been refactored to support this change with some extensibility and type safety tweaks.


steets250 (approved):
Looks good to me. Adding to cart was functional and the localStorage output looks correct 🚀

PR #485, opened 2021-12-08
Proper Google Font Loading

steets250:
We previously only loaded one weight of the font 'DM Sans', which was the 400 weight (normal). No matter what weight was assigned, the browser would only display either the normal font or a slightly bolder version.

I updated the import method to match what Google Fonts suggests as best practice, and cleaned up the weights used. Since we now have 400 (normal), 500 (medium), and 700 (bold), I went through and cleaned up and references as follows:

100, 200, 300, normal => 400
500, 600 => 500
700, 800, bold => 700

I also made sure the linter required numeric font weights, as this makes things clearer going forward.


public/index.html
```diff
@@ -19,6 +19,9 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
+    <link rel="preconnect" href="https://fonts.googleapis.com"> 
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin> 
+    <link href="https://fonts.googleapis.com/css2?family=DM+Sans:wght@400;500;700&display=swap" rel="stylesheet">
```
StormFireFox1, line 24: We _could_ download the fonts locally if we intend to make sure we don't get screwed by Google outages, but it's not a requirement.

StormFireFox1 (approved):
Checked the preview, fonts do look bolded now!

Looks good to me, just left a nit in there.

PR #486, opened 2021-12-14
Checkout Page

### LINKED ISSUE #467: Cart Display
**Name(s)**
`CartDisplay`

**Description**
The cart display component takes in an array of item options + quantities, and displays a cart view. There are two modes that this component has. A writable view should display "Edit" and "Remove" links, allow user to adjust quantity, and remove items from the cart. A read-only view should only display the items, but not allow anything to be adjusted. When "Edit" is clicked, the item option field should turn into a dropdown, and then a different item option can be selected.

**Image**
Screen Shot 2021-09-12 at 11 37 12 AM

**Figma**
https://www.figma.com/file/7i9MK6r5km22gynYlqC0Uj/DesignxPortal?node-id=718%3A6633

**Input Data**
This component takes in `writable`, which is a boolean to indicate the component mode, and an array of `CartItemModel`, which contains a `MerchandiseItemOptionModel` and a quantity. This data will be generated by the `Cart Display`'s parent component and then passed in.

**State**
When an item quantity is changed or an item is removed, the correct event should be dispatched to update the state.

**Notes**
-

### LINKED ISSUE #468: Cart Page
**Name(s)**
`CartPage` and `CartPageContainer`

**Description**
The cart page should process data to be passed to a `CartDisplay` component, and should also give the ability for the user to click a "Checkout" button. This component should fetch cart from state, load each corresponding `MerchandiseItemModel`, and then pass this information to `CartDisplay` as an array of object with the structure `{item: MerchandiseItemModel, quantity: number}`. When "Checkout" is clicked, the cart verification API route should be pinged. If the response is a success, then the user should be directed to the checkout page (`/store/checkout`). If the response contains an error, display it using the `notify()` function, and disable the "Checkout" button until the cart is updated.

**Image**
Cart Page

**Figma**
https://www.figma.com/file/7i9MK6r5km22gynYlqC0Uj/DesignxPortal?node-id=447%3A6674

**Input Data**
There is no input data for this component.

**State**
The component should load the cart from state, and use that to help in processing the data to be passed to `CartDisplay`.

**Notes**
-

yesean:
resolves #467, resolves #468 

This PR implements the `CartPage` and `CartDisplay` components. It also tweaks the cart logic and types to better fit the store workflow.


steets250 (changes requested):
The code looks solid! I didn't have any suggestions to make, and the data seems to be displaying and sending correctly. The only issue that I encountered was that discounts aren't being displayed, so the item costs and totals are incorrect. Item options have a `discountPercentage` field, so the actual cost of an item is `item.price * (1 - item.discountPercentage/100)`. Don't worry about displaying the crossed out original price with the sale price next to it, you can just display the sale price when applicable.

yesean:
> The code looks solid! I didn't have any suggestions to make, and the data seems to be displaying and sending correctly. The only issue that I encountered was that discounts aren't being displayed, so the item costs and totals are incorrect. Item options have a `discountPercentage` field, so the actual cost of an item is `item.price * (1 - item.discountPercentage/100)`. Don't worry about displaying the crossed out original price with the sale price next to it, you can just display the sale price when applicable.

updated it to display discounted prices now

steets250 (approved):
LGTM, feel free to merge 🚀

PR #496, opened 2021-12-21
Added almost all of the fields for create item

trulyronak:
"this will be good for your resume" - @steets250 

"At what cost" - @trulyronak


steets250 (approved):
LGTM 🚀

Issue #428, opened 2021-04-05
Admin: Edit Event Page


steets250:
Add the current event image on the edit event page so that the upload button doesn't imply that no image has been added.

PR #498, opened 2021-12-22
Admin: Edit Event Page 

trulyronak:
Addresses #428

Adds the current event image on the edit event page so that the upload button doesn't imply that no image has been added.


steets250 (approved):
LGTM 🚀

PR #502, opened 2021-12-23
Added the admin prepare page

trulyronak:
Issues (now added in #503)
- `PublicOrderItem` doesn't have
    - an item name (so we use the uuid rn)
    - a correct quantity (it always seems to be -1 in `item.option.quantity`)
    - a user name (so we use the user uuid rn)


steets250 (approved):
LGTM 🚀

PR #504, opened 2021-12-23
Increase Layout Padding

steets250:


trulyronak (approved):
lgtm 🚀 we should add global style variables

PR #505, opened 2021-12-24
Admin Form Validation

steets250:
Adds form validation to the collection add/edit page and the item add/edit page. Correctly accounts for the difference in inputs between an item with and without variants.


trulyronak:
@duanwilliam can you battle test this

steets250:
😳

duanwilliam:
:pika:

trulyronak (changes requested):
Almost there, but images should be required.

src/store/components/AdminItemPage/index.tsx
```diff
@@ -21,30 +22,43 @@ interface AdminItemPageProps {
...
-
-  quantity?: string;
-  price?: string;
-  discountPercentage?: string;
-}
+const AdminItemPageFormSchema = Yup.object().shape({
+  collection: Yup.string().required('Required'),
+  itemName: Yup.string().min(2, 'Too Short').max(50, 'Too Long').required('Required'),
+  description: Yup.string().min(2, 'Too Short').max(1000, 'Too Long').required('Required'),
+  // hidden: Yup.boolean().required('Required'),
```
trulyronak, line 29: should these comments be in here?

src/store/components/AdminItemPage/index.tsx
```diff
@@ -21,30 +22,43 @@ interface AdminItemPageProps {
...
+  monthlyLimit: Yup.number().min(1, 'Too Low').max(10, 'Too High').required('Required'),
+  lifetimeLimit: Yup.number().min(1, 'Too Low').max(10, 'Too High').required('Required'),
+  // hasVariantsEnabled: Yup.boolean().required('Required'),
```
trulyronak, line 32: same wiht this comment, is it needed?

src/store/components/AdminItemPage/index.tsx
```diff
@@ -21,30 +22,43 @@ interface AdminItemPageProps {
...
+    then: Yup.string().min(2, 'Too Short').max(50, 'Too Long').required('Required'),
+  }),
+  options: Yup.array().when('hasVariantsEnabled', {
+    is: (hasVariantsEnabled) => hasVariantsEnabled,
+    then: Yup.array().of(
+      Yup.object().shape({
+        uuid: Yup.string(),
+        value: Yup.string().min(2, 'Too Short').max(50, 'Too Long').required('Required'),
+        price: Yup.number().min(1, 'Too Low').max(1000000, 'Too High').required('Required'),
+        quantity: Yup.number().min(1, 'Too Low').max(1000000, 'Too High').required('Required'),
```
trulyronak, line 44: won't this make it impossible to edit a sold out item? i feel like we should allow quantities of 0 (to signify sold out?)

idk im not sure how it works on the backend after an item is purchased (does quantity get modified?)

trulyronak (changes requested):

src/store/components/AdminCollectionPage/index.tsx
```diff
@@ -1,59 +1,57 @@
...
   collection?: PublicMerchCollection | undefined;
 }
 
 interface AdminCollectionPageForm {
-  title?: string;
-  themeColorHex?: string;
-  description?: string;
+  title: string;
+  themeColorHex: string;
+  description: string;
```
steets250, line 24: Formik requires all fields to have a value at all times (including for initial values), so I removed the optionals and added the type to the `initialValues` object.

src/store/components/AdminCollectionPage/index.tsx
```diff
@@ -1,59 +1,57 @@
...
 const AdminCollectionPage: React.FC<AdminCollectionPageProps> = (props) => {
   const { collection } = props;
   const creatingCollection = !collection;
   const title = creatingCollection ? 'Create Collection' : 'Edit Collection';
 
+  const initialValues: AdminCollectionPageForm = {
+    title: collection?.title ?? '',
+    themeColorHex: collection?.themeColorHex ?? '',
+    description: collection?.description ?? '',
+  };
```
steets250, line 44: `initialValues` was moved outside of the `<Formik>` tag so typing could be applied.

src/store/components/AdminItemPage/index.tsx
```diff
@@ -22,79 +23,115 @@ interface AdminItemPageProps {
...
+          quantity: option.quantity.toString(),
+          discountPercentage: option.discountPercentage.toString(),
+        }))
+      : [],
+    // Variants Disabled
+    quantity: (!item?.hasVariantsEnabled && item?.options[0].quantity?.toString()) || '',
+    price: (!item?.hasVariantsEnabled && item?.options[0].price.toString()) || '',
+    discountPercentage: (!item?.hasVariantsEnabled && item?.options[0].discountPercentage.toString()) || '',
+  };
+
```
steets250, line 121: `initialValues` was moved outside of the `<Formik>` tag so typing could be applied.

src/store/components/AdminItemPage/index.tsx
```diff
@@ -22,79 +23,115 @@ interface AdminItemPageProps {
...
     discountPercentage: string;
   }[];
 
-  quantity?: string;
-  price?: string;
-  discountPercentage?: string;
+  quantity: string;
+  price: string;
+  discountPercentage: string;
 }
```
steets250, line 48: Formik requires all fields to have a value at all times (including for initial values), so I removed the optionals and added the type to the `initialValues` object. The only two fields where this isn't the case are `existingPicture` and `newPicture`, which are uniquely controlled by an external component.

src/store/components/AdminItemPage/index.tsx
steets250: I'm going to address this issue in a separate PR. The way backend handles quantity updates is via a separate route that only allows positively incrementing quantities, so I'm going to ask about changing that, and then will address what the valid field values are.

src/store/components/AdminItemPage/index.tsx
```diff
@@ -22,79 +23,115 @@ interface AdminItemPageProps {
...
-            }
-            return errors;
-          }}
+          initialValues={initialValues}
+          validationSchema={AdminItemPageFormSchema}
           onSubmit={async (values, { setSubmitting }) => {
             const url = creatingItem ? `${Config.API_URL}${Config.routes.store.item}` : `${Config.API_URL}${Config.routes.store.item}/${item?.uuid}`;
 
-            const payload = values.hasVariantsEnabled
+            const payload: any = values.hasVariantsEnabled
```
steets250, line 134: Casting this value to `any`, so that the below `delete` line doesn't require additional typing.

src/store/components/AdminItemPage/index.tsx
```diff
@@ -145,7 +183,7 @@ const AdminItemPage: React.FC<AdminItemPageProps> = (props) => {
 
             if (values.newPicture) {
               const formdata = new FormData();
-              formdata.append('image', (values.newPicture as unknown) as Blob);
+              formdata.append('image', values.newPicture);
```
steets250, line 186: By making `values.newPicture` the type `Blob`, this typing issue has been resolved.

src/store/components/AdminItemPage/index.tsx
trulyronak: damn typescript is annoying :( 

the only way to avoid [`any` here](https://stackoverflow.com/questions/63702057/what-is-the-logic-behind-the-typescript-error-the-operand-of-a-delete-operato) is to create an entire interface for the type of payload that makes `options` optional (but that's not worth the ffort)

trulyronak (approved):
lgtm!

PR #508, opened 2022-01-03
Update loginUser Action

steets250:
If the user attempts to access a portal link directly and isn't signed in, they're redirected to the login page. After logging in, they're taken to the home page, rather than their previously desired destination. The `verifyToken` route now encodes the destination into the login URL, so that upon login, the user is taken to the correct destination.

### DIFF
--- a/src/auth/authActions.ts
+++ b/src/auth/authActions.ts
@@ -51,10 +51,19 @@ export const loginUser: ThunkActionCreator = (values, search) => async (dispatch
       isAdmin: userData.admin,
     });
 
-    // Redirect to home on login.
-    if (search !== '') {
-      dispatch(replace(`/checkin${search}`));
+    const params = new URLSearchParams(search);
+
+    const code = params.get('code');
+    const destination = params.get('destination');
+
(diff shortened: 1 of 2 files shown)


trulyronak:
Bug - steps to reproduce:
- yarn start
- login with acm@ucsd.edu
- click on the admin page
- go to create event
- logout
- go to http://localhost:8080/admin/awardPoints on the same page
- failure

Screen Shot 2022-01-04 at 10 47 18 PM

steets250:
Thanks @trulyronak, that should be fixed now.

StormFireFox1 (commented):
Looks good, but there's a little design change we could consider:

Browser history is already a state tracker for the URL's of the portal, so I figure we can use that to check the previous URL and redirect properly instead of keeping the URL in our Redux reducers.

steets250:
#516

steets250:
@trulyronak I hath resolved the issue you hath presented unto me

trulyronak (approved):
lgtm, thanks for fixing that issue @steets250

PR #509, opened 2022-01-03
Stylelint Upgrade

steets250:
When opening a `.less` file in VSCode with the Stylelint plugin enabled, the following message is given: 

Stylelint version 13.7.2 is no longer supported. While it may continue to work for a while, you may encounter unexpected behavior. Please upgrade to version 14.0.0 or newer. See the migration guide for more information.

https://github.com/stylelint/vscode-stylelint#migrating-from-vscode-stylelint-0xstylelint-13x

I went ahead and upgraded stylelint and its related packages to their latest version. This also entailed some new required packages (`postcss` and `postcss-less`), and resulted in new styling rules being applied.


steets250:
Need to revisit tomorrow; there are some issues where the stylelint upgrade is breaking less.

steets250:
Ok finally fixed

trulyronak (approved):
poggers good styling

PR #511, opened 2022-01-05
Timeframe Simplification

steets250:
This PR aims to simplify the use of timeframes in relation to the leaderboard and past events sections of the site. For the leaderboard, the components and containers have been condensed and simplified. I removed the infinite scroll, so that members could more easily search for their names, and won't have to scroll 100 at a time to find themselves. For the past events, I made some minor styling updates and also updated the way past events are loaded. For both sections, they load the current year's leaderboard and events by default. This means the leaderboard is smaller, and less past events are shown. Unfortunately, checking on the testing backend doesn't allow for full testing, but I'll be able to confirm correct behavior when we deploy to the site and can roll back as needed.


steets250:
@trulyronak for this one there were a lot of code changes that involve some oddly wired functions, feel free to just give the timeframes a sanity check on the deploy preview, and try refreshing / switching pages to ensure there's no noticeable breakage.

src/event/containers/PastEvents.tsx
```diff
@@ -46,12 +42,11 @@ interface PastEventsContainerProps {
...
   fetchAttendance: Function;
   fetchPastEvents: Function;
-  updateTimeframe: Function;
-  timeframe: string;
 }
 
 const PastEventsContainer: React.FC<PastEventsContainerProps> = (props) => {
-  const { auth, events, attendance, fetchAttendance, fetchPastEvents, timeframe, updateTimeframe } = props;
+  const { auth, events, attendance, fetchAttendance, fetchPastEvents } = props;
+  const [timeframe, setTimeframe] = useState(getCurrentYear()?.name ?? 'All Time');
```
trulyronak, line 49: out of scope i think: would be nice to use the similar typing [that `ucsd-quarters-years` uses](https://github.com/acmucsd/ucsd-quarters-years/blob/main/src/bounds.ts#L12) - since we do use that here.

src/leaderboard/leaderboardReducer.ts
```diff
@@ -1,47 +1,23 @@
...
-        users: [],
-        offsetToUsers: new Map(),
-      };
-      return {
-        ...newState,
-        error: action.payload,
-      };
-    }
     case FETCH_LEADERBOARD: {
       // TODO: Look into Immutables.
```
trulyronak, line 12: question: what does this mean/  is it worth looking into this now?

src/event/containers/PastEvents.tsx
trulyronak, line 49: when would `getCurrentYear` fail?

trulyronak:
I understand the reasoning behind wanting to remove infinite scroll - however I wanted clarification on a few cases
- for the "All Time" leaderboard, we have ~2000 members. That's a bit too much to load, what are your thoughts about potentially having a "load more" button rendering at the bottom of the page? (instead of infinite scroll)
- I also noticed that `getCurrentYear` could fail - when would that happen, and is loading the entire 2K+ leaderboard the best alternative to that?

overall seems ok though, I'll approve with some comments to look at

trulyronak (approved):

trulyronak (changes requested):
actually requesting changes here - there should be a better screen shown if there isn't any data

Screen Shot 2022-01-05 at 3 23 57 PM

src/event/containers/PastEvents.tsx
StormFireFox1: When you pass in an academic year not currently in the dataset of the package; you can check out the available academic years [here](https://github.com/acmucsd/ucsd-quarters-years/blob/main/src/data.ts).

trulyronak (approved):
lgtm

PR #512, opened 2022-01-05
Store checkout

paulpan05:
Checkout page implementation. Successful order placement but wasn't able to test successful order redirect due to low account balance (will test later).


src/store/containers/CheckoutPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const getFuturePickup = async (onFailCallback: () => void) => {
+    try {
+      const url = `${Config.API_URL}${Config.routes.store.pickup.future}`;
+      const result = await fetchService(url, 'GET', 'json', {
+        requiresAuthorization: true,
+      });
+      const eventMap = {};
+      result.pickupEvents.forEach((item, index, arr) => {
+        const startDate = parseDate(new Date(item.start));
+        const endDate = parseDate(new Date(item.end));
```
steets250, line 59: When converting the date to a human-readable string, we can use the `moment` library to make things a bit cleaner. For example, `Moment(item.start).format('MMM D, YYYY [@] hh[:]mm a')` would return a string like `Jan 5, 2022 @ 04:08 pm`. This removes the need for the helper enums and function.

src/store/components/CheckoutPage/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+            cart.forEach((item) => {
+              inputData.order.push({ option: item.option.uuid, quantity: item.quantity });
+            });
+            const url = `${Config.API_URL}${Config.routes.store.order}`;
+            try {
+              const result = await fetchService(url, 'POST', 'json', {
+                requiresAuthorization: true,
+                payload: JSON.stringify(inputData),
+              });
+              history.push(`/store/${result.order.uuid}`);
```
steets250, line 61: On success, we want to go to `/store/order/:uuid`.

steets250 (changes requested):
Looks solid overall, left a couple comments regarding date parsing + the order url.

steets250 (approved):
Looks good to me! 🎉

PR #514, opened 2022-01-06
Rename PageLayout Selector

steets250:
There was a class name overlap between the `PageLayout` selector and the profile cart class name, this rename resolves that issue as a quick fix.


trulyronak (approved):
lgtm

PR #520, opened 2022-01-11
Jesus Take the Wheel

steets250:


src/auth/containers/requireStoreAccess.tsx
```diff
@@ -13,10 +13,12 @@ const withStoreAccess = (Component: React.FC) => (props: { [key: string]: any })
...
-    if (state === UserState.PENDING || !(emailDomain === 'ucsd.edu' || emailDomain === 'acmucsd.org')) {
-      redirectHome();
-    } else {
-      setPermitted(true);
+    if (email) {
+      if (state === UserState.PENDING || !(emailDomain === 'ucsd.edu' || emailDomain === 'acmucsd.org')) {
+        redirectHome();
+      } else {
+        setPermitted(true);
+      }
```
StormFireFox1, line 21: How come there's no email? When does that happen?

StormFireFox1 (commented):
Nit, and I'll approve!

src/auth/containers/requireStoreAccess.tsx
steets250: `requireStoreAccess` is nested within `requireAccess`, which ensures that the user is signed in and their data has loaded. By checking that the email is defined, it ensures that we don't throw an invalid email error before the user profile has loaded.

StormFireFox1: Yep, makes sense!

StormFireFox1 (approved):
LGTM!

PR #523, opened 2022-01-11
Checkout Page Updates

steets250:
I added sorting by start date to pickup events, and then gave the checkout page its own style file so I could add a bottom margin. (Since the dropdown is absolutely positioned, there's no way for the page it's in to know how tall it is, so I just always add a bottom margin to avoid cutoff).


trulyronak (approved):
dope, lgtm

PR #532, opened 2022-01-25
Fix Incorrect Date Comparison

steets250:
In order to determine that the current time is too close to the pickup event, the current time should be greater than the pickup start time minus two days, not less than.


StormFireFox1 (approved):
Yeah, that’d do it :))

PR #538, opened 2022-02-12
Update Leaderboard Loading Messages

steets250:
Rather than showing "No users found for this timeframe!" while initially populating leaderboard data, the portal now displays "Loading users...".


StormFireFox1 (approved):
Looks good!

For future thought, we might want to consider adding spinners (or bar that look like content is loading) for components in the portal that do load after a page loads in for the updates.

PR #539, opened 2022-02-15
Auth route rewrite

paulpan05:
Merging requireAuth, requireAdminAuth, and requireStoreAccess components.


src/auth/containers/protectRoute.tsx
```diff
@@ -0,0 +1,116 @@
...
+  authenticated: state.auth.authenticated,
+  pathname: state.router.location.pathname,
+  search: state.router.location.search,
+  isAdmin: state.auth.admin,
+  state: state.auth.profile.state,
+  email: state.auth.profile.email,
+});
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  redirectHome: (rejectRoute: string) => {
```
steets250, line 106: Small nit: can we rename this to something like `redirect` instead of `redirectHome`, since it's now going to a dynamically chosen route?

steets250 (approved):
Looking solid, I this will get cleaner as we review our state management as well. Before this gets merged, can you address the comment and then delete the no longer needed `requireAuth`, `requireAdminAuth`, and `requireStoreAccess` files?

StormFireFox1 (approved):
Looks good! I agree with Steven's nits, essentially.

Issue #462, opened 2021-07-13
Improve State Management


steets250:
The current Redux store divides state into components as follows:

```
admin: AdminReducer,
auth: AuthReducer,
event: EventReducer,
leaderboard: LeaderboardReducer,
profile: ProfileReducer,
store: StoreReducer,
```
This ends up preventing actions in a specific area from updating state in a separate area. We should look into re-organizing the state by data field, rather than exactly mirroring the folder setup. We may also want to consider if Redux is necessary, and if using `useContext` and `useState` is a better option for our use case.

StormFireFox1:
We've decided to try using Redux Toolkit in order to refactor our state. We have notes on this in the board Notion, so you can read about the design choice there.

As for tasks to do, this involves taking all the boilerplate we have and moving it to effectively what RTK does (which is slices).

I've assigned Sean on this, since he's most experienced with RTK, and we can discuss the various changes in the PR he will make when we review it.

- [ ] Create slice files for RTK for each feature folder in UI (so `storeActions, storeReducer` → `storeSlice`)
- [ ]  Compose slice file into separate RTK store
- [ ]  Move actions and reducers to RTK store one at a time(test if portal breaks with each slide adding)
  - If this proves too hard to do without breaking the other parts of portal, just move it all at once and we'll test bugs at review
- [ ]  Remove old Redux store

PR #542, opened 2022-02-18
Portal API Upgrade

### LINKED ISSUE #516: Update Router Navigation
Browser history is already a state tracker for the URL's of the portal, so I figure we can use that to check the previous URL and redirect properly instead of keeping the URL in our Redux reducers.

steets250:
I'll go over what changes this PR entails roughly by commits. Don't be alarmed by file count, there are ~55 changed files that are simply disconnected from Redux. However, even though the file count is high, I tried to minimize the risk to functionality to save that for future, more moderately sized PRs.

## Independent History

The first step involved pulling the history management out of the Redux state. It looks like history was added in to allow for page navigation via Redux dispatches, but that's never something that was implemented and just added to boilerplate code. There's now a `history.ts` file that can be imported wherever a `history.push` or `history.replace` is needed.

## Removal of `ThunkActionCreator`

During the Typescript rewrite, a static type `ThunkActionCreator` was added probably to resolve typing issues. However, what this did was effectively prevent Typescript from caring about the method signature of those functions, so we were unable to verify correct input. I removed this along with the corresponding package since we do not need dynamic dispatched functions anymore.

## Addition of Backend Code

I copied `Enums.ts`, `ApiRequests.ts`, and `ApiResponses.ts` directly from the backend repo. The rationale for doing this was in preparation for having the backend package that will eventually be used. By mocking the existence of the package's files, the frontend can be prepared to wire everything up smoothly, rather than having to contend with this rewrite later on.

## Creation of Backend Functions

There is now an `api` folder, which contains all currently used (nothing related to feedback was brought over) functions, with the function names, inputs, and return types exactly matching the ones used in the backend. These functions aren't currently used and will be something that's implemented more granularly over the next few weeks. Once these functions are implemented, making the switch to an npm package should be seamless. During this creation, the `fetchService` utility was located inside the `api` folder.

## Update Existing Action Implementatino

Dispatch is no longer needed for making `history` calls and is also no longer needed as a parameter, as we can instead make a single import of the Redux store and reference it directly. Because of this, all references to history and dispatching were updated, so all existing `*Action.ts` files were updated. You may notice that some types have been switched to `any`. This is to avoid having to resolve typing issues as a part of this update. During the transition to the new `api` functions, typing issues can be resolved.

## Update Existing Action Usage

Functions no longer need to be wired to a `dispatch`, so they can be imported and used directly in components. This entailed the highest number of file changes, as many calls to the redux `connect` function were no longer needed, and usages of action functions could be referenced directly instead of via `props`. These changes have the highest risk of causing breakages if a non-typed component is expecting a prop that is no longer wired up. This would only be an issue with the older Formik forms (used specifically for admin pages and event administration).

## What's Next

There are a couple of ways to proceed. These changes can be merged into the existing codebase without affecting current behaviors, and then the transition to `api` can occur, or this can remain as a PR and only be merged once the transition to `api` is complete. The latter will result in a cleaner codebase after merge but is more prone to merge conflicts as other changes to the codebase occur.

Please let me know if you notice a cleaner way of doing things, I went through a few revisions of the `api` methods, and am hoping that they are clean.

Also a quick note, the files located in `api` come in handy for the backend when generating the methods to export publicly in the package.

Resolves #516. Ties in with #312 and #462.


steets250:
Oop gonna chunk this into smaller PRs, but I'll leave this up as a draft for future reference.

PR #543, opened 2022-02-19
Independent History

steets250:
History management has now been pulled out of the Redux state. It looks like history was added in to allow for page navigation via Redux dispatches, but that's never something that was implemented and just added to boilerplate code. There's now a `history.ts` file that can be imported wherever a `history.push` or `history.replace` is needed outside of a component. For usages inside of a component, `useHistory` or `useLocation` can continue to be used. This removes the need for the `connected-react-router` package, as our router is now state independent.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
     "antd": "3.24.3",
     "antd-img-crop": "^4.0.2",
     "babel-plugin-import": "^1.12.0",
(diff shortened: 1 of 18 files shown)


StormFireFox1:
Merging this PR with Sean's work, since this is useful for the Redux toolkit refactor.

PR #550, opened 2022-03-11
Redux cleaning

yesean:
This PR is a refactor of our redux usage, most notably the help of [redux-toolkit](https://redux-toolkit.js.org/). The main goal was to simplify the code and reduce the amount of boilerplate brought on by redux. This is achieved in multiple ways:

1. Using toolkit's `createSlice` and `createAsyncThunk` to define redux reducers and actions. These helper functions reduce a lot of redux code (with the help of [immutable-js](https://github.com/immutable-js/immutable-js/) and typescript) which allows all of the redux to be colocated into a single `*Slice.ts` file.
2. Replacing the use of the `connect` HOC with the `useDispatch` and `useSelector` hooks which effectively remove the need for most container components.
3. General purpose refactoring to keep the code (including some redux-adjacent stuff) as DRY but simple as possible.

Addtional notes:
* `*{Reducer,Actions,Types}.ts` files have been replaced with `*Slice.ts` and a `utils.ts`. the slice file holds all of the redux logic and the utils file holds the non-redux logic (fetching, notifying, etc). generally, the functions in utils are imported into the slice to provide any needed redux for the utility. this was done to decouple the redux and non-redux logic as much as possible.
* the `redux_reducers.ts, combineReducers, createStore` setup has been replaced with the simpler `redux/store.ts, configureStore` (the toolkit recommended way)
* redux-toolkit provides excellent typescript support for the store and actions which was previously severely lacking. I tried to type as much as I could but some parts of the store (in particular, the event slice) are poorly typed (dynamically changing slice structures) so it was hard to infer exactly what fields would be in each slice. In any case, the current type safety is far better than before (some fields may still be missing though).
* I needed the `Awaited` utility type which was introduced in Typescript 4.5 so I upgraded the typescript package to latest. Unfortunately, typescript changed their AST in 4.0 but `create-react-app` uses an older, incompatible version of `typescript-eslint/parser` so I used selective dependency resolutions (`resolutions` in `package.json`) to force the new version of the eslint parser. Nevertheless, everything is working fine now, though this change may be better suited for another PR?
* I was pretty aggresive with the refactor and I did change some of the business logic and I'm not too familiar all of the inner workings of the portal so I may have made some wrong assumptions about various processes (auth, checkin, etc). I haven't tested it too extensively, for the most part, it works but I'm hoping this PR can not only serve as a code review but an open call for testing.


StormFireFox1:
This is some great work, @yesean, well done!

I've looked over your PR at a first glance, and overall, it's good! I haven't reviewed the business logic you've changed yet, but I'll probably do that after finals week. I think there's a couple questions that came up while I looked over it we might want to discuss:
- Currently, most of each feature's `utils.ts` is effectively the API calls, whereas conventionally utilities were simply local functions we used to simplify business logic. Currently, @steets250 was moving API calls under an API folder of sorts, maybe we wanna do something similar here.
- Also, on a more fun note, like you said, I don't think we need containers anymore! It's kinda hard to truly move away from them, considering how prominent they are all around the code, so I suggest we rename all our containers to `pages`, which makes more sense as to what purpose they serve; representing each page from our portal.

What do y'all think of this?

steets250:
Ah sorry I've been MIA for the past week(s) about this, you can completely ignore the `api-updates` branch, that was my wild adventure of "oh, now that this is fixed I can also fix this", which I'm still in the process of splitting up into more reasonably sized PRs. Tomorrow I'll take review this, and hopefully the switch to separate api functions can just happen in place once these changes are either merged in, or a PR can be opened against this branch instead if we want to hold off on merging until everything is done.

src/auth/components/EmailVerificationPage/index.tsx
```diff
@@ -1,8 +1,8 @@
 import React, { useEffect, useState } from 'react';
 import { Button } from 'antd';
 import { useParams, useHistory } from 'react-router-dom';
-import { verifyEmail } from '../../authActions';
 import './style.less';
+import { verifyEmail } from '../../utils';
```
steets250, line 5: Can we keep `import './style.less';` as the last import?

src/auth/components/ResendEmailVerificationPage/index.tsx
```diff
@@ -1,8 +1,8 @@
 import React, { useEffect, useState } from 'react';
 import { Button } from 'antd';
 import { useHistory } from 'react-router-dom';
-import { sendEmailVerification } from '../../authActions';
 import './style.less';
+import { sendEmailVerification } from '../../utils';
```
steets250, line 5: See above comment.

src/auth/containers/requireAdminAuth.tsx
```diff
@@ -1,50 +1,32 @@
 import React, { useEffect } from 'react';
```
steets250, line 1: The changes in this, `requireAuth`, and `requireStoreAccess` will need to be re-applied to Paul's new access protection component. We'll want to merge from master to incorporate those changes.

src/layout/components/NavBarHorizontal/index.tsx
```diff
@@ -23,20 +23,20 @@ const NavBarHorizontal: React.FC<NavBarHorizontalProps> = (props) => {
...
+          <NavTileItem icon={ProfileIcon} />
         </NavLink>
         <NavLink activeClassName="selected" to="/discord">
-          <NavTileItem icon={DiscordIcon} text="Discord" />
+          <NavTileItem icon={DiscordIcon} />
         </NavLink>
         {isAdmin && (
           <NavLink activeClassName="selected" to="/admin">
-            <NavTileItem icon={AdminIcon} text="Admin" />
+            <NavTileItem icon={AdminIcon} />
```
steets250, line 39: How do these changes tie in to redux stuff?

src/store/localStorage.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const loadCart = () => {
+  try {
+    const serializedCart = localStorage.getItem('cart');
+    if (serializedCart === null) return {};
+
+    const cart = JSON.parse(serializedCart);
+    return cart as Cart;
+  } catch (err) {
+    return {};
```
steets250, line 11: If we console.log for store, shouldn't' we do the same for load? And perhaps it should be a `notify` so that the user sees that where was an error and it's not a mystery why something didn't save/load.

src/admin/adminSlice.ts
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable no-param-reassign */
```
StormFireFox1, line 1: Why is this required?

src/admin/components/AddAttendanceForm/index.tsx
```diff
@@ -1,14 +1,11 @@
-import React, { useState, ChangeEventHandler, FocusEventHandler, FormEventHandler } from 'react';
-import { AutoComplete, Checkbox, Form, Input, Button, Tooltip, Tag, Icon, Select } from 'antd';
+import { AutoComplete, Button, Checkbox, Form, Icon, Input, Select, Tag, Tooltip } from 'antd';
+import React, { FormEventHandler, useState } from 'react';
```
StormFireFox1, line 2: Typically, I think in React the 'react' import should always be first in file (unless you're in Next.js, where you don't need it). I might wrong on this, and CRA might have removed that need. We should readjust that in files where it happens.

StormFireFox1 (changes requested):
Left some nits for any things Steven didn't mention that I thought would be important but otherwise, before we merge, I feel like we should still discuss the items I mentioned in my first comment before we merge.

src/auth/components/EmailVerificationPage/index.tsx
yesean: For every file that I modified, I ran Typescript's built in `organizeImports` command from the language server `tsserver`. I'm not sure if we can configure it, but I like having an automated, opinionated formatter for imports.

src/auth/components/ResendEmailVerificationPage/index.tsx
yesean: same as above

src/layout/components/NavBarHorizontal/index.tsx
yesean: The `text` prop was unused in the `NavTileItem` component so I removed it. I think it came up when I was resolving lint warnings.

src/store/localStorage.ts
yesean: Good catch, I was planning on removing the log and doing what `loadCart` does. But I like the idea of adding a `notify` so something like:
```tsx
try {
// ...
} catch (err) {
	notify(err);
	return {};
}
// ...
```
Does that sound good?

src/admin/adminSlice.ts
yesean: The `no-param-reassign` rule in our eslint config has the option `{props: true}` (from airbnb) which forbids mutating properties of parameters to reduce side effects. However, since `redux-toolkit` uses immer for `createSlice`, it's more idiomatic to mutate the `state` argument when defining reducers since it'll be equivalent to copying the parameter and is generally easier to read as well. I still think it's a good rule, so I've just disabled it in all slice files since createSlice will usually mutate the state parameter.

`redux-toolkit` has a [discussion](https://github.com/reduxjs/redux-toolkit/issues/521) on this which tldr: airbnb is too restrictive, just use `{props: false}` which is the eslint default anyway. 

Thoughts?

src/admin/components/AddAttendanceForm/index.tsx
yesean: [same as above](https://github.com/acmucsd/membership-portal-ui/pull/550#discussion_r837275349). I don't believe the react import needs to come first, but it's more idiomatic. Also, I think you're referring to the [new JSX Transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) where the `React` import is not needed and JSX completely handled at compile time. Unless we need to use React 16.3, we could upgrade to 16.4 or 17 to take advantage of it. Facebook even provides a [codemod](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports) to automatically remove the unneeded imports. All we'd have to do is upgrade `react` and `react-scripts` I believe.

yesean:
> This is some great work, @yesean, well done!
> 
> I've looked over your PR at a first glance, and overall, it's good! I haven't reviewed the business logic you've changed yet, but I'll probably do that after finals week. I think there's a couple questions that came up while I looked over it we might want to discuss:
> 
> * Currently, most of each feature's `utils.ts` is effectively the API calls, whereas conventionally utilities were simply local functions we used to simplify business logic. Currently, @steets250 was moving API calls under an API folder of sorts, maybe we wanna do something similar here.

I don't really see the distinction between API calls and business logic as API calls feel like another step in the logic. Moreover, the calls are generally localized to a section of the store and not shared between sections, so to me, it makes sense to keep them at a section-level rather than the top-level. Perhaps, `utils` isn't best place for API calls but I still think they should be localized rather than a top level API folder. Of course, if a backend API package is introduced, this may all be irrelevant.

> * Also, on a more fun note, like you said, I don't think we need containers anymore! It's kinda hard to truly move away from them, considering how prominent they are all around the code, so I suggest we rename all our containers to `pages`, which makes more sense as to what purpose they serve; representing each page from our portal.

`pages` sound good to me, glad to be moving away from containers :grinning: 

> What do y'all think of this?

src/admin/containers/CreateEventForm.tsx
```diff
@@ -64,4 +63,4 @@ const FormikCreateEventForm = withFormik({
   },
 })(CreateEventForm as React.FC);
 
-export default connect(null, { postEvent, copyLink })(FormikCreateEventForm);
+export default connect(null, { postEvent })(FormikCreateEventForm);
```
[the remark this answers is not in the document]
yesean: The only instances where I left `connect` was with Formik since there isn't a way to access the `useDispatch` or `useSelector` hooks with the `withFormik` HOC. In order to remove the `connect` usage there, we'd have to rewrite the Formik code with Formik hooks or different component APIs which is probably better suited for another PR.

src/admin/components/EditEventPage/index.tsx
```diff
@@ -19,7 +19,7 @@ interface EditEventPageProps {
 const EditEventPage: React.FC<EditEventPageProps> = (props) => {
   const { event } = props;
 
-  const EditEventCustomForm = (EditEventForm as unknown) as React.ComponentClass<any>;
+  const EditEventCustomForm = EditEventForm as unknown as React.ComponentClass<any>;
```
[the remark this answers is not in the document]
yesean: That must've come from `prettier`, I don't recall manually editing that. `as` is probably left associative so the parenthesis don't matter.

StormFireFox1:
Just to recap all the discussion from above and meetings:

- [ ] Add ESLint plugin for organizing imports
- [ ] Handle merge conflicts for Paul's PR
- [ ] Disable ESLint rule for `no-param-reassign`

yesean:
> Just to recap all the discussion from above and meetings:
> 
> * [ ]  Add ESLint plugin for organizing imports
> * [ ]  Handle merge conflicts for Paul's PR
> * [ ]  Disable ESLint rule for `no-param-reassign`

Fixed

StormFireFox1 (approved):
_Totally_ looks good to me! 🚀 👀

PR #553, opened 2022-03-22
fix spacing

farisashai:
before

Screen Shot 2022-03-21 at 11 46 17 PM

after

Screen Shot 2022-03-21 at 11 46 33 PM


leestanley:
Should these two pieces of text be put in a `display: flex` with `justify-content: space-between` instead? They are currently in a span where the `.experience` class is in a `float: right`

https://github.com/acmucsd/membership-portal-ui/blob/7f0ccee4bafd93e7d7ec6899d460c76c19362524/src/profile/components/ProfileCard/style.less#L36-L38

michl1001 (approved):
Looks good, other responsiveness issues but likely requires a larger PR

steets250 (commented):
Looks good to me, thanks for addressing this.

PR #555, opened 2022-03-28
Colors and dark mode

paulpan05:
Changes were made to complete dark mode implementation. Also, some linting issues were fixed.


src/admin/components/EditEventForm/style.less
```diff
@@ -74,9 +74,9 @@
...
 
   .delete-button {
-    background-color: @red;
-    border-color: @red;
-    color: @white;
+    background-color: var(--theme-danger-1);
+    border-color: var(--theme-danger-1);
+    color: var(--theme-background);
     font-size: @general-body;
     font-weight: 500;
```
StormFireFox1, line 81: nit: We use 500 and 700 a lot in the codebase as weights for normal and bold text. I figure we can make those variables. Maybe `--normal-weight/--bold-weight` or `--normal-font-weight/--bold-font-weight`. Up to you, though.

StormFireFox1 (changes requested):
Since it looks like you made a new PR off of Trevor's branch, I only reviewed your last 5 commits (including merges) to make sure we didn't do redundant reviews.

Honestly, looks great code-wise! There is only one nit for the code, and I've already commented it.

I did just look over the deploy preview, and realized the sun icon for the display mode toggle is full yellow. This is kind of hard to see in light mode, and the low contrast is not very accessible. This is what I'm talking about:

The display mode toggle with a yellow sun icon on a white background.

If we could change up the color to be gray, I think it'd be great!

StormFireFox1:
@paulpan05 I'll get the icon changed and we can merge this PR; I think it's fine to leave the font weights for now, we'll look into replacing them later for the Ant Design stuff. Ideally we merge this before Kickoff.

steets250 (approved):
LGTM 🚀

StormFireFox1 (approved):
We'll skip over the nit for now, we can address it later.

PR #562, opened 2022-04-28
fix external url redirect and breakpoint

farisashai:
This PR fixes two separate bugs.

1: External URLs are not always prefixed with https:// so they sometimes will not properly redirect. Fixed with existing utility function to deal with this.
Screen Shot 2022-04-27 at 8 16 18 PM

2: The breakpoint at 768px was inconsistent on the navbar and resulted in two shifts at 767px and 768px rather than a single shift.


StormFireFox1 (approved):
Looks good to me! 🚀

PR #567, opened 2022-05-12
Fix Password Reset API Call

### LINKED ISSUE #566: Reset Password Broken
Clicking the submit button on the page to set a new password (https://members.acmucsd.com/resetPassword/:hash) is non-functional, and doesn't seem to be making a network call at all.

steets250:
There was an issue where there was a stray double function, so calling the helper function to reset a user's password was only generating a function and wasn't actually executing the call.

Resolves #566


StormFireFox1 (approved):
Nice catch!

PR #569, opened 2022-05-18
Update Store Admin Pickup Dropdown Conditions

steets250:
Previously, store admins were only shown currently ongoing and future pickup events in the dropdown for store fulfillment. However, this meant that a pickup event disappeared as soon as it was over, so if it hadn't already been marked as complete, then someone from dev needed to find its uuid and manually access the page. The dropdown condition is now changed to display both past and future pickup events that have a status of active, which will allow admins to both fulfill orders and mark pickup events as complete even after their original time window has passed.

### DIFF
--- a/src/store/containers/AdminFulfillPage.tsx
+++ b/src/store/containers/AdminFulfillPage.tsx
@@ -5,7 +5,7 @@ import { useAppDispatch } from '../../redux/store';
 import { PublicOrderPickupEvent } from '../../types';
 import { notify } from '../../utils';
 import AdminFulfillPage from '../components/AdminFulfillPage';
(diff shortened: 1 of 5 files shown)


steets250:
@StormFireFox1 if u wanna nit the fact that we have the same process of fetching past pickup events, future pickup events, and then filtering the same way in 3 files, I can work on cleaning that up into a helper function. I followed how it was done now and didn't feel like doing cleanup as part of this PR 😂

StormFireFox1:
Yeah, we can descope the cleanup for another PR.

StormFireFox1 (approved):
Awesome! 🚀

PR #571, opened 2022-05-31
Fix error handling

yesean:
This PR adds `UserError` and `AuthError` types to differentiate between server errors. In the redux cleaning, I was dispatching logout on all server errors but now I only dispatch logout on `AuthError`. 

In the future, I think there should be a more unified way of handling `UserError` as generally, we just notify with an error message which can be abstracted. I've already created this kind of abstraction in `src/auth/authSlice.ts` in the `withLogout` HOF which takes an async function `f` and returns an async thunk that dispatches a logout when `f` throws an `AuthError`. I think a similar HOF could be created for `UserError` which sends a notify if a `UserError` is thrown. If such a `withNotify` HOF is created, then `withLogout` would probably have to be modified to allow for error chaining before creating the async thunk.

### DIFF
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -3,6 +3,7 @@ import { showNotification } from '@mantine/notifications';
 import Storage from './storage';
 import DiamondDisplay from './store/components/DiamondDisplay';
 import { FetchServiceOptions, HttpRequestMethod, MimeType, OrderStatus, PublicMerchItemOption } from './types';
+import { AuthError, UserError } from './errors';
 
 export const notify = (title: string, description: string) => {
   showNotification({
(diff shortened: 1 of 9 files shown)


src/utils.tsx
```diff
@@ -187,9 +188,9 @@ export const fetchService = async (url: string, requestMethod: HttpRequestMethod
   });
 
   const { status } = response;
-  if (status === 401 || status === 403) onFailCallback?.();
+  if (status === 401 || status === 403) throw new AuthError('');
```
StormFireFox1, line 191: Would it be doable to add some kind of message in here for the AuthError? I figure maybe the error field of the response might return something useful?

PR #576, opened 2022-08-16
Added indicator that displays when there are no upcoming events

raymosun:


sunouray:
Not sure how to format this. From here, the linter seems to want the div to be less indented
image

But like this, the linter wants it to be more indented
image

farisashai (approved):
Looks good - this is a really nice improvement from the previous version!

raymosun:
No events indicator looks like this 
image

PR #584, opened 2022-09-30
Fix #578 dashboard css responsiveness

### LINKED ISSUE #578: Fix CSS Responsiveness on different viewpoints for dashboard cards
Screen Shot 2022-09-27 at 11 59 30 AM

raymosun:
Cards now have a minimum width and flex into 2 lines instead of overflowing.

### DIFF
--- a/src/profile/components/ProfileCard/style.less
+++ b/src/profile/components/ProfileCard/style.less
@@ -12,13 +12,20 @@
   justify-content: center;
   margin: 0;
   max-width: @db-card-width;
+  min-width: @db-card-width-min;
   overflow: hidden;
   white-space: nowrap;
+
+  .ant-card-body {
+    max-width:100%
+  }
 
   h2 {
     font-size: @card-heading-1;
     font-weight: 700;
     margin: 0;
+    text-overflow: ellipsis;
+    overflow: hidden;
   }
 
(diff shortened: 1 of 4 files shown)


src/profile/components/ProfileCard/style.less
```diff
@@ -41,6 +48,7 @@
 
   .info {
     display: inline-block;
+    max-width: calc(100% - 145px);
```
farisashai, line 51: Nitpick but a good convention for accessibility reasons is using [rem length values in CSS](https://www.freecodecamp.org/news/what-is-rem-in-css/). 1rem = 16px so 10rem = 160px and it's usually fine to round to the nearest whole rem rather than converting it exactly. Also wondering if you can explain what the 145px hardcoded value is for

farisashai (changes requested):
Responsiveness looks and feels way better 👍 Left one small nitpick to fix. Can you also include a screenshot of the visual changes in the PR before I approve

src/profile/components/ProfileCard/style.less
raymosun: Also edited some screenshots into the original comment 👍

farisashai (approved):
looks good!

PR #587, opened 2022-10-19
Reverted Redux Cleaning and subsequent commits

trevorkw7:
Redux Cleaning (#550) fails to run at production while containing too many file changes to individually bug fix and address. Thus, a decision was made to revert this commit and all commits after this commit was made.


farisashai (approved):
Looks good, I'm not seeing any of the bugs we currently have on the latest repo.

Issue #579, opened 2022-09-27
Optional Facebook Link is Required to Create Events


farisashai:
![Screen Shot 2022-09-27 at 12 00 45 PM](https://user-images.githubusercontent.com/33165426/192613083-dc2e4898-76f0-43f3-9578-5ead8b781f4f.jpg)

Trying to submit an event to the membership portal without a facebook link fails the form validation but the field says it's optional.

Either remove this validation if it's actually optional or mark it as required to avoid confusion.

PR #590, opened 2022-10-27
Changed create and edit event facebook field to indicate not optional (#579)

trevorkw7:
Removed "(optional)" form the Facebook event link title to reflect that the Facebook event link is required to pass form validation.

# Before
image

# After
image

### DIFF
--- a/src/admin/components/CreateEventForm/index.tsx
+++ b/src/admin/components/CreateEventForm/index.tsx
@@ -94,7 +94,7 @@ const CreateEventForm: React.FC<CreateEventFormProps> = (props) => {
               <p className="form-error">{errors.pointValue ? errors.pointValue : null}</p>
             </Form.Item>
           </div>
-          <Form.Item label="Facebook Event Link">
+          <Form.Item label="Facebook Event Link (Optional)">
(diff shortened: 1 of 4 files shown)


farisashai (commented):
This fixes the confusion on the UI right now but can we confirm the actual constraints we're placing on the user.

Does the portal API actually require this field or do we have unnecessary client-side validation right now?

trevorkw7:
> #

The portal API doesn't require this field in the POST request but client-side validation is still happening because most admins creating events agree that creating a Facebook event for each event in the membership portal is a good practice.

farisashai:
> The portal API doesn't require this field in the POST request but client-side validation is still happening because most admins creating events agree that creating a Facebook event for each event in the membership portal is a good practice.

I kinda feel like this should be a standard enforced by our marketing team when they're filling out new events in that case rather than a mandatory front-end field if it's not actually required. I can imagine that not every single ACM event requires an event link so I'd rather allow this flexibility for them than potentially burden them more and make someone enter a fake nonsense URL to complete the submission.

trevorkw7:
Makes sense to me. I'll update this pull request with a new commit to the frontend form validation to remove the requirement for a URL to be present in the field.

farisashai (approved):
LGTM 🪨

PR #591, opened 2022-11-04
Rewrite Retroactive Attendance Input Field

farisashai:
Rewrote the retroactive attendance admin panel input component to make it faster to add multiple attendees and prevent Formik from resetting the client-side state internally and making the UI buggy.

### Before
Screenshot 2022-11-03 at 7 58 00 PM

### After
Screenshot 2022-11-03 at 7 57 39 PM


raymosun (approved):
Just wondering if the "all attendees are staff" functionality from the original page was no longer required. 
image
Otherwise, lgtm 👍

PR #592, opened 2022-11-11
Redo dashboard card responsiveness

raymosun:
Redo #584 after the rollback. Cards now have a minimum width and flex into 2 lines instead of overflowing.
image


farisashai (approved):
Looks good! 👍

PR #593, opened 2022-11-11
Update auto-assign workflow

farisashai:
Modified the Github action workflow to assign current Portal UI members as reviewers for each other's code and assign the PR to its creator.


raymosun (approved):
Makes sense looks good 👍

PR #599, opened 2023-01-17
Resume upload user side

raymosun:
Work in progress PR. Resume upload user side is basically done, but not fully tested. Also might want to change how updating resume visibility works.

image
image
image


farisashai (changes requested):
I know this is still in progress but I'm going to leave some nitpicks on functionality from demoing the page and future TODOs to be aware of.

**1. "Resume not visible to recruiters" doesn't make sense to show if there's no resume.**

Screenshot 2023-01-26 at 7 30 44 PM

**2. The toggle input doesn't have enough space between the file upload above or between the switch and the text. Try to play around with this and see if we can make it look a bit better and maybe more consistent with other spacing values on the page.**

Screenshot 2023-01-26 at 7 32 02 PM

3. We still need to fix the visibility body field on the initial POST request from the new backend change.
Screenshot 2023-01-26 at 7 40 17 PM

src/profile/components/ManageResume/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const [currResumeSharing, setCurrResumeSharing] = useState(false);
+  const [uploadState, setUploadState] = useState('none');
+  const [uploaded, setUploaded] = useState(false);
+  const [onCooldown, setOnCooldown] = useState(false);
+
+  useEffect(() => {
+    setCurrResumeSharing(!!user.profile.resumes?.[0]?.isResumeVisible);
+  }, [user]);
+
+  const uploadIcon = (
```
trevorkw7, line 40: Might want to try putting these in assets and import them as React Components

trevorkw7:
Everything else looks good to me in terms of functionality. The only behavior I saw during my testing that might need to be changed is how the spinning indicator for an upload continues to spin infinitely even after we get a status 500 error and toast message for a file being too large. 
Screen Shot 2023-02-09 at 1 13 55 PM

raymosun:
> Everything else looks good to me in terms of functionality. The only behavior I saw during my testing that might need to be changed is how the spinning indicator for an upload continues to spin infinitely even after we get a status 500 error and toast message for a file being too large. Screen Shot 2023-02-09 at 1 13 55 PM

Fixed the points you brought up in latest commits, thanks!

farisashai (approved):

Issue #577, opened 2022-09-27
Fix Communities on Event Creation


farisashai:
We should remove Hack, Innovate, and Design as official community types for events from the portal to maintain consistent branding across our platforms.

![Screen Shot 2022-09-27 at 11 57 28 AM](https://user-images.githubusercontent.com/33165426/192612488-5f86ad46-f8c8-45cc-a02b-8652b9ab4f79.jpg)

The main website pulls events from the portal and no longer supports styling for hack, design, and innovate so they are missing styles and appear with black text.

![Screen Shot 2022-09-27 at 11 52 49 AM](https://user-images.githubusercontent.com/33165426/192611629-965317b0-57f8-44cb-a9a3-5505fb8ec520.jpg)

PR #602, opened 2023-01-27
Fix Communities on Event Creation

Jvil26:
Resolves issue #577 

Screenshot 2023-01-26 at 12 44 59 AM


farisashai (approved):
looks good to me 🚀 

In the future, it might be nice to have a constant in our config file that stores a list of all ACM orgs and which ones are active to reduce these kind of changes if our board structure is modified.
```
const config = {
  allOrgs: ['General', 'AI', 'Cyber', 'Design', 'Hack', 'Innovate']
  activeOrgs: ['General', 'AI', 'Cyber']
}

```

PR #604, opened 2023-01-28
Improve authorization for hidden store items

### LINKED ISSUE #580: Hidden store items need stronger front-end authorizations
**Hidden items should not be viewable to non-admins.** 
If you enter the link to a hidden item on the ACM store, the page is visible when it should just redirect back to the main dashboard or some homepage instead.

Screen Shot 2022-09-27 at 12 05 29 PM

Screen Shot 2022-09-27 at 12 05 39 PM

**Hidden should not be able to be inserted into a cart even as an admin**
We already have the back-end check that refuses this transaction. We should match this standard on the front-end.
Screen Shot 2022-09-27 at 12 07 27 PM

Jvil26:
Resolves Issue #580

Screenshot 2023-01-26 at 12 45 20 AM

### DIFF
--- a/src/store/components/ItemPage/index.tsx
+++ b/src/store/components/ItemPage/index.tsx
@@ -1,5 +1,7 @@
 import React, { useState } from 'react';
 import { connect } from 'react-redux';
+import { Dispatch } from 'redux';
+import { replace } from 'connected-react-router';
 import { Modal } from 'antd';
 
 import { PublicMerchItemWithPurchaseLimits, PublicMerchItemOption } from '../../../types';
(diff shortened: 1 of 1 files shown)


src/store/components/ItemPage/index.tsx
```diff
@@ -150,4 +158,15 @@ const ItemPage: React.FC<ItemPageProps> = (props) => {
...
-export default connect(null, { addToCart })(ItemPage);
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  redirect: (rejectRoute: string) => {
+    dispatch(replace(rejectRoute));
+  },
+  addToCart,
+});
+
+const mapStateToProps = (state: { [key: string]: any }) => ({
+  isAdmin: state.auth.admin,
```
farisashai, line 169: I double checked the portal account role specifications and it looks like store administrator roles also need full admin access only on the store UI so we have to change the logic to include `store_manager` too. 

Let's describe the logic better by renaming the variable to something like `isStoreAdmin` or `canViewHiddenItems` to demonstrate what the boolean is for more accurately

Full Description of Roles here if you want to understand the different scenarios better: https://github.com/acmucsd/membership-portal/blob/master/README.md#testing-information

raymosun (approved):
lgtm! 🚀

farisashai (approved):
looks good 🚀

Issue #582, opened 2022-09-28
Buggy Store Merch Editing Interface


farisashai:
Lots of fun bugs to fix 😃 

Steps to reproduce:

- View merch store as an admin
- Click the edit button on an item
- Try to file upload a new image
**- Bug: Upload freezes and does not let you change the image.**
![Screen Shot 2022-09-28 at 5 25 16 AM](https://user-images.githubusercontent.com/33165426/192777907-23e18cac-b000-41b2-b69a-2927caae1816.jpg)
- Try to delete file upload.
**- Bug: Page crashes and goes blank.**
- Refresh page
**- Bug: Collection Name field displays the wrong value**
![Screen Shot 2022-09-28 at 5 24 44 AM](https://user-images.githubusercontent.com/33165426/192777834-3cae7e6a-ff78-4456-9fee-002c77d9d531.jpg)

- Try to add variants to an existing item
**- Bug: Disabled unless we completely delete the item and create a new one**

![Screen Shot 2022-09-28 at 5 25 58 AM](https://user-images.githubusercontent.com/33165426/192778033-d867a88b-7a41-4158-a66f-eac497c81bbb.jpg)

- Apply a 100% discount code.
**- Bug: Allows the discount code to be applied in the editing item UI and backend but doesn't actually do anything. (Disable this value if it's not allowed)**
- ![Screen Shot 2022-09-28 at 5 23 12 AM](https://user-images.githubusercontent.com/33165426/192777560-6a052a6c-35a9-4286-ab71-995760f1cc64.jpg)
![Screen Shot 2022-09-28 at 5 23 20 AM](https://user-images.githubusercontent.com/33165426/192777585-75f1cc93-ebbc-4c88-a643-708cf44fc6c3.jpg)

UI Nitpick:
I'm not a big fan of this quantity adjustment field as a way of changing the quantity.

We should either be able to directly enter the quantity we want to set the item to or maybe have "+" and "-" increment buttons to the side of the quantity field to manually adjust the value and save it

![Screen Shot 2022-09-28 at 5 28 41 AM](https://user-images.githubusercontent.com/33165426/192778562-abb016ab-c2b5-4cd2-8189-30f05ce54246.jpg)

PR #605, opened 2023-02-02
Fix Buggy Store Merch Editing Interface #582

raymosun:
Fixing issues in Buggy Store Merch Editing Interface #582


src/store/components/AdminItemPage/index.tsx
```diff
@@ -129,6 +130,12 @@ const AdminItemPage: React.FC<AdminItemPageProps> = (props) => {
     discountPercentage: (!item?.hasVariantsEnabled && item?.options[0].discountPercentage.toString()) || '',
   };
 
+  // converts uuid of a collection to { label: ..., value: ... }
+  const getOptionFormat: (string) => Option = (uuid: string) => {
```
farisashai, line 134: Fixed with a better long-term solution in #608, if we revert this code back, we can merge as a solution to the other bug being resolved.

PR #607, opened 2023-02-13
Fix Add to Cart Broken Functionality

farisashai:
Fixed bug preventing add to cart functionality from working due to issue in incorrect mapDispatchToprops function because addToCart was undefined.

Expected behavior to test for:
- Add to Cart works correctly now
- Hidden item url redirect for non-store admins (Previous PR change)


shravanhariharan2 (approved):
Functionality seems to work for me, not sure if my approval can merge the PR but adding it just in case it does

image

Issue #600, opened 2023-01-21
Store Admin Collection Title Display


farisashai:
If you click edit item on the store admin interface, the collection will sometimes displays as the UUID instead of the collection title. It’s really weird and nconsistent, you might have to refresh the page a few times to replicate the bug.

![Screenshot 2023-01-21 at 12 12 31 PM](https://user-images.githubusercontent.com/33165426/213885427-bcb6f6f5-0f0c-4bca-98f8-dc9669cf1cf6.jpg)

PR #608, opened 2023-02-18
Fixed admin edit collection UUID bug #600

ryan-truong:
Fixed issue when admin edited item in store the collection title would show the uuid instead of the title of the collection #600

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "less": "^3.9.0",
     "less-loader": "^5.0.0",
     "lodash": "^4.17.15",
-    "moment": "^2.29.2",
+    "moment": "^2.29.4",
     "prop-types": "^15.6.2",
     "react": "^16.13.1",
     "react-app-rewired": "^2.1.3",
--- a/src/admin/containers/CreateEventForm.tsx
+++ b/src/admin/containers/CreateEventForm.tsx
@@ -10,7 +10,7 @@ import { isURL } from '../../utils';
 const CreateEventSchema = Yup.object().shape({
   title: Yup.string().required('Required'),
   location: Yup.string().required('Required'),
-  eventLink: Yup.string(),
+  eventLink: Yup.string().nullable(),
   pointValue: Yup.number().required('Required').moreThan(0, 'Must be greater than 0').integer('Must be an integer'),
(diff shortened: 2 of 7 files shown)


farisashai (changes requested):
Let's try to double check the full usage of the variable before we change it. There might be a better approach to changing the display name visually while keeping the uuid.

bug: If we edit a collection, don't make any edits, and hit save, the behavior would fail with this code because the collection uuid is how we tell the backend which collection we are editing and we would no longer have access to the uuid if we remove it. 

Screenshot 2023-03-03 at 5 04 27 PM

farisashai (changes requested):
I'm still gettting the bug on your branch preview
Screenshot 2023-04-04 at 2 45 55 PM

I'll try to explain the convo we had again yesterday for your future understanding. We have two separate functions `fetchCollectionsFunction` and `fetchItemFunction` in this file that run asynchronously (return a Promise bc we don't deterministically know how long they'll take to complete). The two ways of handling an async function are 

Old Solution: Promise Chaining
```JavaScript
fetchCollectionsFunction()
    .then(() => {
        // do something after the async function has returned
    })
    .catch(error => {
        // handle errors
    })
```
By chaining the `.then()` function, we tell our code to execute all lines in the file fully and remember to run the code in the `then` block after the function has returned something.

Newer Solution: Async/Await
```JavaScript
async () => {
   try {
    const response = await fetchCollectionsFunction();
      // do something after the async function has returned
    } catch (error) {
      // handle errors
    }
}
```
By including the keyword `await` in the second example, we tell the function to pause execution until the function returns a value.

_Our Bug_
Since our code is running two independent async functions with promise chaining, 
```JavaScript
fetchCollectionsFunction().then(() => do something)
fetchItemFunction().then(() => do something)
```
We are essentially starting two timers at the exact same time and telling each one what to do after they finish so the order of the functions in our code won't make a difference. We need to tell our code precisely which function should run first and which one should run second. We can do this by placing one of the function calls inside the `then` block of the other function 
```JavaScript
fetchCollectionsFunction().then(() => {
  fetchItemFunction().then(() => {})
})
```

farisashai (approved):
Looks good, thank you for your patience!

PR #609, opened 2023-03-31
Filters still not working

WishingWell13:
Admin Resume Table Sorting in progress


farisashai (changes requested):
Which part of the code are you stuck on right now? It looks like all of the filters are working properly in the deployment preview and there are no errors in the console.

Also, it looks like your code is failing the linting check on our codebase right now. That means that the code quality isn't consistent with our style rules. We use ESLint in the terminal to check linting for us in the project but I would highly recommend you install the VSCode extension for ESLint for syntax error highlighting as you code.

If you install it, you'll see these red lines on your code, 

Screenshot 2023-04-01 at 5 16 54 PM

Hovering over any red lines will give you this pop up where you can click quick fix and fix all auto-fixable problems to resolve many of the issues for you automatically. If there are any remaining red areas, you'll need to read the error messages and ensure you can fix it.

Screenshot 2023-04-01 at 5 17 27 PM
Screenshot 2023-04-01 at 5 17 39 PM

.env.development
```diff
@@ -1,2 +1,3 @@
+//REACT_APP_API_URL=https://api.acmucsd.com
```
farisashai, line 1: delete this commented line

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -13,16 +13,40 @@ interface UserResumeTableProps {
 
 const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
   const { resumes } = props;
+  const [yearFilters, setYearFilters] = useState([{ text: '2026', value: '2026' }]);
```
farisashai, line 16: Can we figure out a better initial value to avoid hardcoding a year like this? By default, we shouldn't have any filters applied to the table when the page loads - how would we represent this in our state?

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -13,16 +13,40 @@ interface UserResumeTableProps {
...
   const [selectedRowKeys, setSelectedRowKeys] = useState<React.Key[]>([]);
 
+  const updateYearFilters = () => {
+    setYearFilters(
+      resumes
+        .filter((value: UserResume, index: number) => {
+          // Resumes.indexOf(value) === index checks for first occurance
+          return resumes.indexOf(value) === index;
```
farisashai, line 24: Single-line arrow functions can be inlined to save space and readability.

Example

From
```JavaScript
mylist.map((element) => {
   return element * 2
})
```

To
```JavaScript
mylist.map(element => element * 2)
```
Note: The thing on the right side of the arrow is the return value. If you need any additional lines of code in the body of the function, this won't work

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -13,16 +13,40 @@ interface UserResumeTableProps {
...
+        })
+        .sort()
+        .map((element) => ({ text: `${element.user.graduationYear}`, value: `${element.user.graduationYear}` })),
```
farisashai, line 27: Can we format this object to be more readable? (Start keys on a new line)
```suggestion
        .map((element) => ({ 
            text: `${element.user.graduationYear}`, 
            value: `${element.user.graduationYear}` 
        })),
```

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -13,16 +13,40 @@ interface UserResumeTableProps {
...
+
   const fileName = 'acm_resumes';
 
+  // const categoriesOfYear = props.resumes.filter((value : UserResume, index : number) =>
+  //     {
+  //         //Resumes.indexOf(value) === index checks for first occurance
+  //         return resumes.indexOf(value) === index;
+  //     }).sort().
+  //     map(element => ({text: `${element.user.graduationYear}`, value: `${element.user.graduationYear}`}));
+
```
farisashai, line 39: Delete unused code

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -13,16 +13,40 @@ interface UserResumeTableProps {
...
   const resumeSelection = {
     onChange: (newSelectedRowKeys) => {
       setSelectedRowKeys(newSelectedRowKeys);
     },
   };
 
+  // const handleChange = (pagination, filters, sorter) => {
+  //   console.log('Various parameters', pagination, filters, sorter);
+  // };
```
farisashai, line 48: Delete

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -53,6 +77,7 @@ const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
     saveZip(fileName, selectedResumeLinks);
   };
 
+  // TODO: Look for sorter key inside documentation
```
farisashai, line 80: Delete once this is completed

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -79,11 +110,28 @@ const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
     {
       title: 'Planned Graduation',
       dataIndex: 'user.graduationYear',
+
+      filterMultiple: true,
+      filters: yearFilters, // [{ text: "2016", value: "2016" }],
+
+      onFilterDropdownVisibleChange: (visible) => {
```
farisashai, line 117: Is this function necessary? If it is, can you explain why?

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -79,11 +110,28 @@ const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
...
+        if (visible) {
+          setTimeout(() => updateYearFilters());
+        }
+      },
+
+      // Value we are filtering by, record = content of current row
+      onFilter: (value: string, record: UserResume) => `${record.user.graduationYear}` === value,
```
farisashai, line 124: Since the graduation year type is a number, does it make sense to check string equality here?

We probably want to check numeric equality of `record.user.graduationYear === value`, however looking at this line of code above,
```
filters: yearFilters, // [{ text: "2016", value: "2016" }],
```
we can't do numeric equality unless we set value to be the **number** 2016 rather than the string. 

yearFilters should be something like 
```
[{
    "text": "2016",
    "value": 2016
}]
```

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -79,11 +110,28 @@ const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
...
     },
     {
       title: 'Date Uploaded',
       dataIndex: 'lastUpdated',
       render: (date: string) => <span>{reformatDate(date)}</span>,
+
+      onFilter: (value, record) => record.name.indexOf(value) === 0,
+      sorter: (a, b) => {
+        return a.lastUpdated.localeCompare(b.lastUpdated);
```
farisashai, line 133: We can inline this function's return value

farisashai: We should also use proper variable names that are representative of the data like (user1, user2) or resumeEntry, etc for whatever this is rather than `a, b`

src/admin/containers/UserResumesTable.tsx
```diff
@@ -10,13 +10,162 @@ interface UserResumesTableContainerProps {
...
+// ];
+
 const UserResumesTableContainer: React.FC<UserResumesTableContainerProps> = (props) => {
   const { resumes, getAllVisibleResumes } = props;
 
   useEffect(() => {
     getAllVisibleResumes();
   }, [getAllVisibleResumes]);
 
+  // Can set resumes to testingData to use testingData
```
farisashai, line 168: We can delete all of the comments in this file and unused code. We just need comments when we're documenting a hard to read portion of code or function for other developers to maintain in the future.

farisashai (changes requested):
Good job on your first issue on for the portal! Everything seems functional so far, I left some comments to ensure we maintain our code quality.

src/admin/components/UserResumeTable/index.tsx
WishingWell13: ESLint edited it to be like so. Let me know if this is OK.
```
resumes
        .filter((value: UserResume, index: number) => resumes.indexOf(value) === index)
        .sort()
        .map((element) => ({ text: `${element.user.graduationYear}`, value: `${element.user.graduationYear}` })),
    );
```

src/admin/components/UserResumeTable/index.tsx
WishingWell13: onFilterDropdownVisibleChange was there to delay the getting of resumes so they would not return null

src/admin/components/UserResumeTable/index.tsx
WishingWell13: I believe ColumnProps only takes a string. I get errors when trying to make it a number. If this is not the case please let me know and I'll spend some time on it.

image

src/admin/components/UserResumeTable/index.tsx
farisashai: I'm going to reopen this comment so I can take a look at it in the codebase later. There should be a cleaner approach to fetch the resumes or potentially have a null check on the resumes being defined but we can go with this if I don't come up with any other suggestions

src/admin/components/UserResumeTable/index.tsx
farisashai:
```suggestion
      onFilter: (value: string, record: UserResume) => record.user.graduationYear === value,
```

I explained the small tweak we need above in a different comment for the value key in setYearFilters. We don't need to modify the columns in any way for this filter check. If you can get the previous line of code working correctly, this should also work in a cleaner way.

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -79,11 +99,26 @@ const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
...
+
+      // Value we are filtering by, record = content of current row
+      onFilter: (value: string, record: UserResume) => `${record.user.graduationYear}` === value,
     },
     {
       title: 'Date Uploaded',
       dataIndex: 'lastUpdated',
       render: (date: string) => <span>{reformatDate(date)}</span>,
+
+      onFilter: (value, record) => record.name.indexOf(value) === 0,
```
farisashai, line 120: What is this line of code used for? I don't think we have any filter applying to the "Date Uploaded" column.

src/admin/containers/UserResumesTable.tsx
```diff
@@ -10,6 +10,154 @@ interface UserResumesTableContainerProps {
...
+//       firstName: 'Olivia',
+//       lastName: 'Rivera',
+//       profilePicture: null,
+//       graduationYear: 2029,
+//       major: 'Bioinformatics',
+//       bio: null,
+//       points: 0,
+//     },
+//   },
+// ];
```
farisashai, line 159: nitpick: Delete this comment

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -13,8 +13,22 @@ interface UserResumeTableProps {
 
 const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
   const { resumes } = props;
+  const [yearFilters, setYearFilters] = useState([{ text: '', value: '' }]);
```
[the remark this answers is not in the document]
WishingWell13: image

This was the error message I was getting when I make it an empty array:

image

image

src/admin/containers/UserResumesTable.tsx
WishingWell13: Bump

src/admin/components/UserResumeTable/index.tsx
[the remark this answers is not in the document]
farisashai: Maybe try typing the variable explicitly in the setState like this
```JavaScript
interface FilterType = { text: string; value: string; }[]
const [yearFilters, setYearFilters] = useState<FilterType>([])
```

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -79,11 +109,24 @@ const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
...
+      },
+
+      // Value we are filtering by, record = content of current row
+      onFilter: (value: string, record: UserResume) => `${record.user.graduationYear}` === value,
     },
     {
       title: 'Date Uploaded',
       dataIndex: 'lastUpdated',
       render: (date: string) => <span>{reformatDate(date)}</span>,
+      sorter: (user1, user2) => user1.lastUpdated.localeCompare(user2.lastUpdated),
```
farisashai, line 129:
```suggestion
      sorter: (user1, user2) => new Date(user1.lastUpdated) < new Date(user2.lastUpdated)
,
```

farisashai (approved):
everything looks good to me 💵

PR #611, opened 2023-04-05
Fix bug where store item variants are not sorted 

farisashai:
Bug was caused by admin page mapping variant options directly without matching the position from the metadata options


alexzhang1618 (approved):
looks good to me 🤓

PR #614, opened 2023-04-07
Fix route naming consistency

farisashai:
Renames client-side routes to match email template from the backend.

Since these routes can only be hit from a direct email link, they aren't linked anywhere else in the codebase that I need to tweak.


alexzhang1618 (approved):
so epic 😍

PR #617, opened 2023-04-25
Andy add sorting to admin resume table

WishingWell13:
Trying to get netlify preview


WishingWell13:
Needed to filter repeats by year, not entire object

CUsersSmithwickDocumentsCSE30HW2
farisashai, line 1: delete this file with your hw 😱 no ai violations here

WishingWell13:
instead of `resumes
  ?.filter((value: UserResume, index: number) => resumes.indexOf(value) === index)`
  do
  
  `.filter(
          (value: UserResume, index: number) => resumes.map((resume) => resume.user.graduationYear).indexOf(value.user.graduationYear) === index,`

src/admin/components/UserResumeTable/index.tsx
```diff
@@ -26,8 +26,10 @@ const UserResumeTable: React.FC<UserResumeTableProps> = (props) => {
     setYearFilters(
       // Resumes.indexOf(value) === index checks for first occurance
       resumes
-        ?.filter((value: UserResume, index: number) => resumes.indexOf(value) === index)
-        .sort((resume1, resume2) => resume1.user.graduationYear - resume2.user.graduationYear)
+        ?.sort((resume1, resume2) => resume1.user.graduationYear - resume2.user.graduationYear)
+        .filter(
+          (value: UserResume, index: number) => resumes.map((resume) => resume.user.graduationYear).indexOf(value.user.graduationYear) === index,
```
farisashai, line 31: Good job finding the bug! On second look over this section to code, i have a couple thoughts i didn't notice earlier.

If you look at what we're doing here, we keep repeatedly pulling the field `resume.user.graduationYear` out of every object when all we really need is **a sorted list of unique graduation years that exists within this list**. A simpler way of doing this with less operations and computational power would look something like this.

Since we don't require any other fields out of the resumes list, using the Set class this way is an effective way of getting unique elements in a list, and we can just throw away all keys that are unused by mapping to the grad year list straight away.
1. Get the full nested list of graduation years contained in this array **once**.
2. Find all unique elements in this list, our set of dropdown options.
3. Sort the years in order.

This would look like
1. `const gradYears = resumes.map(resume => resume.user.graduationYear)`
2. `const uniqueYears = [...new Set(gradYears)]`
3. `const dropdownOptions = uniqueYears.sort((a, b) => a - b)`

farisashai (changes requested):
looks good to address the bug but I left some commments showing how we can make the code a bit safer and cleaner 😮 lmk if you have questions

src/admin/components/UserResumeTable/index.tsx
WishingWell13: Why would the full nested list of graduation years be called once? Wouldn't it be called each time we need to update/setYearFilters? And if it is outside of the function, then doesn't it risk populating off of an empty list?

farisashai: Oh oops yeah I meant once per updating the filters is simpler than calling filter, map, and sort on an array of full objects with custom functions passed in everywhere.

farisashai (approved):
looks much better!

PR #627, opened 2023-12-31
make compatible with multiple store gallery images

alexzhang1618:
- hello 
- i am not on call for this

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -78,7 +78,12 @@ export interface PublicMerchItem {
   uuid: Uuid;
   itemName: string;
   collection?: PublicMerchCollection;
-  picture: string;
+  merchPhotos: {
+    uuid: string;
+    uploadedPhoto: string;
+    uploadedAt: string;
+    position: number;
+  };
   description: string;
   monthlyLimit: number;
   lifetimeLimit: number;
--- a/src/types.ts
+++ b/src/types.ts
@@ -83,7 +83,7 @@ export interface PublicMerchItem {
     uploadedPhoto: string;
     uploadedAt: string;
     position: number;
-  };
+  }[];
   description: string;
   monthlyLimit: number;
   lifetimeLimit: number;
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { notification } from 'antd';
 
 import Storage from './storage';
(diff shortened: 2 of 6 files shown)


src/types.ts
```diff
@@ -78,7 +78,12 @@ export interface PublicMerchItem {
   uuid: Uuid;
   itemName: string;
   collection?: PublicMerchCollection;
-  picture: string;
+  merchPhotos: {
```
SheepTester, line 81: Why is this not an array

src/types.ts
```diff
@@ -78,7 +78,12 @@ export interface PublicMerchItem {
...
+    uuid: string;
+    uploadedPhoto: string;
+    uploadedAt: string;
+    position: number;
+  };
```
SheepTester, line 86: make this an array:

```suggestion
  merchPhotos: {
    uuid: string;
    uploadedPhoto: string;
    uploadedAt: string;
    position: number;
  }[];
```

it type checked because the `noImplicitAny` TypeScript option was disabled, but it meant that `merchPhotos[0]?.uploadedPhoto` had type `any`

SheepTester (changes requested):

src/types.ts
alexzhang1618: fixed oops

src/utils.tsx
```diff
@@ -309,3 +309,15 @@ export const parseOrderStatus = (status: OrderStatus) => {
...
 };
+
+/**
+ * Given a merch item, return the first picture associated with it.
+ */
+export const getDefaultMerchItemPicture = (item: PublicMerchItem | undefined): string | undefined => {
+  if (item && item.merchPhotos.length !== 0) {
+    // Get the item here with the smallest position (since it doesn't always arrive sorted)
+    item.merchPhotos.sort((a, b) => a.position - b.position);
+    return item.merchPhotos[0].uploadedPhoto;
```
SheepTester, line 320: since you only need the image with the lowest position, sorting it is unnecessary

but whatever, this is frontend webdev, performance is an afterthought

SheepTester (approved):
lessgoo

src/utils.tsx
alexzhang1618: yeah its at max 5 times I think we will survive o7

PR #628, opened 2024-01-01
Multiple Gallery Store Images: Add cart backwards compatibility + default images

alexzhang1618:
- Fixing bug where store order item images weren't showing properly
- Adding backwards compatibility for old carts containing items before the MGSI migration
  - Accepts a picture if it's defined in the cart
  - More type-safety for merchOrders
  - Also returns a default image if none of those things work


farisashai (approved):
LGTM 🚀

steets250 (approved):
Tested using the deploy preview and no longer seeing an error with the cart 👍

